PR #12, opened 2020-03-04
Move Istio reconciler code from knative/serving & preserving history.

tcnghia:
We will follow-up with moving files and renaming directories.


chaodaiG:
omg, 1732 files

mattmoor:
lol 3800 commits.  Let's see if we can get a filtered commit history that isn't a fork of Serving.  Some things to look at are `git clone --filter`, or maybe worth investigating what `k8s-publishing-bot` does.

That said, you should time box this.  Preserving history pre-split isn't make or break.  I'd frame how much time you are spending figuring this out against some theoretical bound on how much time it might save folks to have the history here (vs. just going to serving's history).

My $0.02, fwiw.

PR #32, opened 2020-03-08
Download test_images and conformance tests.

tcnghia:


hack/update-deps.sh
```diff
@@ -44,15 +44,17 @@ done
 readonly DEP_FLAGS
 
 # Ensure we have everything we need under vendor/
-dep ensure ${DEP_FLAGS[@]}
+set +e
+for try in $(seq 1 5); do
```
mattmoor: what is this? 🤣

tcnghia: magic!  that week GitHub constantly times out.

just realize it has a bug.

PR #39, opened 2020-03-12
Start testing with Istio 1.5.

tcnghia:


third_party/istio-latest
```diff
@@ -1 +1 @@
-istio-1.4.4
\ No newline at end of file
+istio-1.5.0/
```
mattmoor: Add trailing newline:
```suggestion
istio-1.5.0/

```

ZhiminXiang: do we need to include `\` ?

tcnghia: This is a symlink, the content is automatically generated by git I think.

PR #46, opened 2020-03-16
Fix name for istio canonical service names.

trshafer:
I accidentally copied the wrong label name. The label name is "canonical-name" not "canonical-service":

https://docs.google.com/document/d/1EsKpYNW55-LqG08zdjp1tAd224T9IEG3CEnusmL-Joc/edit#heading=h.lxsagz3b8886.

I have asked about exporting the label strings from istio side. Currently it's exported on pilot, and I don't think this should have a dependency on pilot: https://github.com/istio/istio/blob/2d1b8d6fc2a30787d397770e6f6aa1eec6a46b03/pilot/pkg/model/service.go#L151

### DIFF
--- a/pkg/defaults/deployment_defaults.go
+++ b/pkg/defaults/deployment_defaults.go
@@ -29,13 +29,11 @@ const (
 	// GroupName is the group name for istio labels and annotations
 	GroupName = "service.istio.io"
 
-	// RevisionLabelKey is the label key attached to k8s resources to indicate
-	// which Revision triggered their creation.
+	// RevisionLabelKey is the label key to define the canonical revision name used by istio.
 	RevisionLabelKey = GroupName + "/canonical-revision"
 
-	// ServiceLabelKey is the label key attached to a Route and Configuration indicating by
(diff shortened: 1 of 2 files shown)


pkg/defaults/deployment_defaults.go
```diff
@@ -29,13 +29,11 @@ const (
...
 
-	// RevisionLabelKey is the label key attached to k8s resources to indicate
-	// which Revision triggered their creation.
+	// RevisionLabelKey is the label key to define the canonical revision name used by istio.
 	RevisionLabelKey = GroupName + "/canonical-revision"
 
-	// ServiceLabelKey is the label key attached to a Route and Configuration indicating by
-	// which Service they are created.
-	ServiceLabelKey = GroupName + "/canonical-service"
+	// ServiceLabelKey is the label key to define the canonical service name used by istio.
```
vagababov: would this be a backward compatible change?

trshafer: No, the old was was just wrong.

PR #48, opened 2020-03-18
Update istio client library to 1.5.0

trshafer:


vagababov:
We tried that in serving and it blew up error rates across the board.
Why do we want it here?
/cc @tcnghia

tcnghia:
@trshafer this update the Istio API client library, not Istio.  Can you please update the PR?  

We are currently testing with Istio 1.5.0 starting https://github.com/knative/net-istio/pull/39 .  Is that what you are trying to achieve, or is there something in the API client library that you want to pull.

Anyway, this vendor client hasn't been used in this repo yet (we're still using the one vendored from serving).  It's only my plate to do that first before we change this.

trshafer:
@tcnghia updated the PR title and commit.

The original rationale to update istio was to address https://github.com/knative/net-istio/issues/47. The referenced istio PR isn't in 1.5.0, but I thought I would go ahead anyway and update the dep.

I don't have any direct need to update the dep right now.

tcnghia:
I see, so let's hold on to this for now if you don't have immediate need.  We can revive after I switch net-istio to use its generated client instead of using Serving's.

PR #49, opened 2020-03-18
Remove the cross namespace OwnerReference

ZhiminXiang:
1. Remove the cross-namespace OwnerReference added when copying secret from user's application namespace to Istio gateway namespace
2. Explicitly delete the secret resources when Ingress is being deleted.

https://github.com/knative/serving/issues/7305

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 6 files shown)


pkg/reconciler/ingress/ingress.go
```diff
@@ -265,6 +262,33 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ing *v1alpha1.Ingress) pk
...
+			return err
+		}
+		for _, nameNamespace := range nameNamespaces {
+			secrets, err := r.GetSecretLister().Secrets(nameNamespace.Namespace).List(labels.SelectorFromSet(
+				resources.MakeTargetSecretLabels(tls.SecretName, tls.SecretNamespace)))
+			if err != nil {
+				return err
+			}
+			for _, secret := range secrets {
+				if err := r.GetKubeClient().CoreV1().Secrets(secret.Namespace).Delete(secret.Name, &metav1.DeleteOptions{}); err != nil {
```
vagababov: Do we want to continue deleting?

ZhiminXiang: I think we may want. Updated the code to try deleting the rest of secrets when error happens.

PR #60, opened 2020-04-05
Clean up unused functions

ZhiminXiang:


ZhiminXiang:
looks like we may need this function for Gateway split work. I will close it for now.

PR #69, opened 2020-04-19
ingress: Skip gateway servers with mutual TLS configured

nairb774:
When mutual TLS is configured on the gateway, a probe is attempted, but is rejected because the caller does not have a client certificate present. With the `knative-ingress-gateway` configured like the following:

```yaml
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*'
    port:
      name: https
      number: 443
      protocol: HTTPS
    tls:
      credentialName: knative-ingress-gateway-tls
      minProtocolVersion: TLSV1_3
      mode: MUTUAL
```

The v0.13.2 networking-istio operator dumps log lines like the following:

```
{
    "level": "error",
    "ts": "2020-04-19T00:47:41.053Z",
    "logger": "istiocontroller.istio-ingress-controller.status-manager",
    "caller": "status/status.go:378",
    "msg": "Probing of https://hello-world-hello-world.ai1.cluster.k8s.local:443 failed, IP: 10.11.57.191:443, ready: false, error: error roundtripping https://hello-world-hello-world.ai1.cluster.k8s.local:443: remote error: tls: alert(116) (depth: 0)",
    "commit": "aa470a6",
    "knative.dev/controller": "istio-ingress-controller",
    "stacktrace": "knative.dev/serving/pkg/network/status.(*Prober).processWorkItem\n\tknative.dev/serving/pkg/network/status/status.go:378\nknative.dev/serving/pkg/network/status.(*Prober).Start.func1\n\tknative.dev/serving/pkg/network/status/status.go:276"
}
```

TLS alert 116 is [`certificate_required`](https://tools.ietf.org/id/draft-ietf-tls-tls13-25.html#rfc.appendix.B.2).

The change is to detect when a gateway is setup for mutual TLS and skip attempting to probe the gateway. In the example above, this will result in 0 probe targets being generated, which is reasonable as this isn't a common or likely "supported" configuration. Regardless, the changes included allow the system to enter a ready state rather than being stuck with `IngressNotConfigured` forever (even though traffic works).

### DIFF
--- a/pkg/reconciler/ingress/lister.go
+++ b/pkg/reconciler/ingress/lister.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 
 	"go.uber.org/zap"
+	istiov1alpha3 "istio.io/api/networking/v1alpha3"
 	"istio.io/client-go/pkg/apis/networking/v1alpha3"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -139,6 +140,10 @@ func (l *gatewayPodTargetLister) listGatewayTargets(gateway *v1alpha3.Gateway) (
 			}
 			tURL.Scheme = "http"
 		case "HTTPS":
+			if server.GetTls().GetMode() == istiov1alpha3.Server_TLSOptions_MUTUAL {
+				l.logger.Infof("Skipping Server %q because HTTPS with TLS mode MUTUAL is not supported", server.Port.Name)
+				continue
+			}
 			tURL.Scheme = "https"
 		default:
 			l.logger.Infof("Skipping Server %q because protocol %q is not supported", server.Port.Name, server.Port.Protocol)
--- a/pkg/reconciler/ingress/lister_test.go
+++ b/pkg/reconciler/ingress/lister_test.go
@@ -431,13 +431,6 @@ func TestListProbeTargets(t *testing.T) {
 							Number:   8080,
 							Protocol: "HTTP",
 						},
-					}, {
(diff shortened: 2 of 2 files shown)


nairb774:
/test pull-knative-net-istio-latest

Looks like a unrelated failure (maybe) - trying again.

JRBANCEL:
/lgtm
🚀

pkg/reconciler/ingress/lister.go
```diff
@@ -139,6 +140,10 @@ func (l *gatewayPodTargetLister) listGatewayTargets(gateway *v1alpha3.Gateway) (
 			}
 			tURL.Scheme = "http"
 		case "HTTPS":
+			if server.GetTls().GetMode() == istiov1alpha3.Server_TLSOptions_MUTUAL {
+				l.logger.Infof("Skipping Server %q because HTTPS with TLS mode MUTUAL is not supported", server.Port.Name)
```
vagababov, line 144: I'd log something other than the port name?

JRBANCEL: There isn't much else to log.
A Gateway CRd looks like:
```
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*'
    port:
      name: http-standard
      number: 80
      protocol: HTTP
  - hosts:
    - yolo1.com
    port:
      name: http-yolo-1
      number: 80
      protocol: HTTP
```

JRBANCEL: Plus it is consistent with the current code, see below code.

PR #71, opened 2020-04-24
Also delete VirtualServices created before Knative 0.12

JRBANCEL:
This is to fix the issue that arises if someone migrates from 0.11- to 0.13+.

In Knative 0.13, the way the name of one of the `VirtualService` is derived from the `Ingress` name changed from `{IngressName}` to `{IngressName}-ingress`. The controller is cleaning the old `VirtualService`, see [code](https://github.com/knative/net-istio/blame/master/pkg/reconciler/ingress/ingress.go#L226), but this is done with the label `networking.internal.knative.dev/ingress` which was introduced in 0.12.

As a result, there are two `VirtualService` matching the same requests. Because the way Istio creates the Envoy configuration is not deterministic, which rule is first (and therefore matching) is random. This means that for each `VirtualService` change, there is a 50% chance that the new `VirtualService` is used and "it works" and a 50% chance that the old `VirtualService` is used and then the behavior varies: if the revision is still around, the traffic will go to that revision, if it is not, it will lead to a 503.

The fix would be to also use the label `serving.knative.dev/route` in the selection logic, this way it would handle `VirtualService` created before 0.12.

/cc @tcnghia @Cynocracy @vagababov

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -34,6 +34,7 @@ import (
 	"knative.dev/net-istio/pkg/reconciler/ingress/resources"
 	"knative.dev/serving/pkg/apis/networking"
 	"knative.dev/serving/pkg/apis/networking/v1alpha1"
+	"knative.dev/serving/pkg/apis/serving"
 	"knative.dev/serving/pkg/network"
 	"knative.dev/serving/pkg/network/status"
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/ingress/ingress.go
```diff
@@ -224,28 +225,34 @@ func (r *Reconciler) reconcileVirtualServices(ctx context.Context, ing *v1alpha1
 	}
 
 	// Now, remove the extra ones.
-	vses, err := r.virtualServiceLister.VirtualServices(ing.GetNamespace()).List(
-		labels.SelectorFromSet(labels.Set{networking.IngressLabelKey: ing.GetName()}))
-	if err != nil {
-		return fmt.Errorf("failed to get VirtualServices: %w", err)
+	selectors := map[string]string{
+		networking.IngressLabelKey: ing.GetName(), // VS created since 0.12
```
Cynocracy, line 229: // VS created between 0.12 and 0.13?

JRBANCEL:
The selector is incorrect, it should be `ing.GetLabels()[serving.RouteLabelKey]`

pkg/reconciler/ingress/ingress.go
JRBANCEL: No, it is from 0.12 on

pkg/reconciler/ingress/ingress.go
```diff
@@ -224,28 +225,34 @@ func (r *Reconciler) reconcileVirtualServices(ctx context.Context, ing *v1alpha1
...
+			labels.SelectorFromSet(labels.Set{k: v}))
+		if err != nil {
+			return fmt.Errorf("failed to get VirtualServices: %w", err)
+		}
 
-	// Sort the virtual services by their name to get a stable deletion order.
-	sort.Slice(vses, func(i, j int) bool {
-		return vses[i].Name < vses[j].Name
-	})
+		// Sort the virtual services by their name to get a stable deletion order.
```
vagababov, line 239: Does this matter here?

pkg/reconciler/ingress/ingress.go
vagababov, line 229: So those VSs have both labels set?

pkg/reconciler/ingress/ingress.go
JRBANCEL: I haven't changed any of this code, it is just inside a loop now

pkg/reconciler/ingress/ingress.go
JRBANCEL: For now, yes

Cynocracy: Why are we deleting them all then? Don't we only want to delete ones from < 0.14?

JRBANCEL:
> The selector is incorrect, it should be `ing.GetLabels()[serving.RouteLabelKey]`

It shouldn't matter in practice because they are the same, but still.

Cynocracy:
@googlebot 

/check-cla

PR #75, opened 2020-04-29
[master] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign tcnghia
/cc tcnghia


vagababov:
some dependency changed, we probably need to pin gomod
@tcnghia

PR #78, opened 2020-04-30
Pin back to knative/{serving,pkg}@release-0.14.

tcnghia:
This is a rollback of https://github.com/knative/net-istio/pull/72  which may have caused new test flakes.


tcnghia:
/assign @ZhiminXiang 

this is to run overnight to confirm if serving dep was the source of test flakes.  I'm running out of ideas :D .

PR #85, opened 2020-05-04
Wait for static IP address to be up and reachable in test.

tcnghia:


test/e2e-common.sh
```diff
@@ -67,6 +67,36 @@ function test_setup() {
...
+  local attempts=100
+
+  STATIC_IP=""
+  for try in $(seq 1 $attempts); do
+    STATIC_IP=$(kubectl get svc $svc -n $ns -ojsonpath={.status.loadBalancer.ingress[0].ip})
+    STATUS_CODE=""
+    if [[ $STATIC_IP != "" ]]; then
+      echo ">> Probing $STATIC_IP"
+      STATUS_CODE=$(curl -s -o /dev/null -w "%{http_code}" http://$STATIC_IP)
+      if [[ $STATUS_CODE != "" && $STATUS_CODE != "000" ]]; then
```
vagababov, line 88: do we consider 5xx also success?

tcnghia: Yes, this is only testing TCP connection to confirm that the underlying network setup is done.  Any non-zero HTTP status code is good indicator here.

PR #87, opened 2020-05-05
Use wait function from knative.dev/test-infra

tcnghia:


chizhg:
The PR merged in test-infra has a bug - https://github.com/knative/test-infra/pull/2010#discussion_r419820466, not sure if it's related to the error here though.

PR #98, opened 2020-05-14
Remove unneeded vendored test files in order to avoid vegeta dep

tcnghia:


hack/update-deps.sh
```diff
@@ -63,6 +63,7 @@ rm -rf $(find vendor/ -name 'OWNERS')
 # Remove unit tests & e2e tests.
 rm -rf $(find vendor/ -path '*/pkg/*_test.go')
 rm -rf $(find vendor/ -path '*/e2e/*_test.go')
+rm -rf $(find vendor/ -path '*/test/e2e/*.go')
```
vagababov, line 66: how about */test/performance and */test/conformance/api ?

tcnghia: We don't pull those into vendor currently

PR #114, opened 2020-05-27
Allow unknown fields in VirtualService and Gateway deserialization

tcnghia:
Fixes https://github.com/knative/serving/issues/7320.


tcnghia:
We can't add a e2e test here for the code change, since we our Istio installation has the Istio webhook (to detect bad-writes from our own code).

PR #121, opened 2020-05-29
Create a shared Gateway for Ingresses that share a wildcard certificate

ZhiminXiang:
We need to handle wildcard certificate specifically because of the Istio limitation https://istio.io/docs/ops/common-problems/network-issues/#404-errors-occur-when-multiple-gateways-configured-with-same-tls-certificate when browser leverages HTTP/2 connection reuse.

This is to fix: https://github.com/knative/serving/issues/7495


pkg/reconciler/ingress/ingress.go
```diff
@@ -148,13 +150,30 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
...
+			// For Ingress TLS referencing wildcard certificates, we reconcile a separate Gateway
+			// that will be shared by other Ingresses that reference the
+			// same wildcard host. We need to handle wildcard certificate specially because Istio does
+			// not fully support multiple TLS Servers (or Gateways) share the same certificate.
+			// https://istio.io/docs/ops/common-problems/network-issues/
+			desiredWildcardGateways, err := resources.MakeWildcardGateways(ctx, wildcardSecrets, r.svcLister)
+			if err != nil {
+				return err
+			}
+			if err := r.reconcileWildcardGateways(ctx, desiredWildcardGateways); err != nil {
```
nak3, line 173: This is a code in the ingres reconcile process. So if only Gateway is manually updated, this function is not called and it is not reconciled.
I mean fox example - gateway is not reonciled until Kingres is refreshed:

```
$ kubectl get gateway -n bar
NAME               AGE
wildcard-df913fc   23s

$ kubectl delete gateway -n bar --all
gateway.networking.istio.io "wildcard-df913fc" deleted

$ kubectl get gateway -n bar
No resources found.

# Gateway is not created until one of Kingress is refreshed or new one is created.
```

nak3: But it is same with non-wildcard gateway, so I wonder if we should fix it or not.

ZhiminXiang: Good question.
For the global gateway under knative-serving namespace, the Gateway change won't trigger Ingress reconcilation because it may cause the situation that one Ingress updates the Gateway which cause the reconcilation of all Ingresses, which is a kind of expensive. In addtion, I will work on migrate the global Gateway model to one-Gateway-per-Ingress model. Once we have this model, the change of a Gateway belonging to one Ingress will trigger the reconcilation of that Ingress.

For the wildcard Gateway, I think it makes sense to trigger the reconcilation of Ingresses when the Gateway is changed or deleted as it won't have the situation of the global Gateway. I added tracker to track the Ingresses associated with the wildcard Gateway. Now manually deleted Gateway will be recreated.

pkg/reconciler/ingress/ingress.go
```diff
@@ -76,8 +77,10 @@ type Reconciler struct {
 	virtualServiceLister istiolisters.VirtualServiceLister
 	gatewayLister        istiolisters.GatewayLister
 	secretLister         corev1listers.SecretLister
+	svcLister            corev1listers.ServiceLister
 
 	tracker   tracker.Interface
+	gwTracker tracker.Interface
```
nak3, line 83: It is difficult for me to understand what this value is for :thinking:  Is it necessary?

nak3:
LGTM except for one nit. 

(leave without lgtm label for @tcnghia and other reviewers.)

pkg/reconciler/ingress/ingress.go
ZhiminXiang: good catch. it is not necessary. deleted it.

PR #125, opened 2020-05-31
Add dummy.go to make config/ importable

chizhg:
Add dummy.go to make config/ importable

/cc @tcnghia @ZhiminXiang


ZhiminXiang:
@chizhg could you please share more context here? Do we need this in order to import yaml files under config/ to knative/serving?

Issue #50, opened 2020-03-20
Investigate and switch to Istio v1beta1 API


tcnghia:
We should take a look at the new Istio v1beta1 API

https://github.com/istio/api/commits/master/networking/v1beta1

if it is ready for use, we should start a feature track to migrate to Istio v1beta1.

howardjohn:
Its the same API with a new label for `beta`. We don't have any conversion webhook or anything so they are identical

tcnghia:
@howardjohn when do you plan to turn down the v1alpha3 API?  I'd love to wait until 1.4 is phased out, so both supported Istio versions (1.5, 1.6) speak v1beta1.

For existing  VirtualServices, would it be ok to have a short time when both v1beta1 VirtualService and v1alpha3 VirtualService exist?  I am paranoid about removing the v1alpha3 VirtualService before the corresponding v1beta1 VirtualService is working properly.

howardjohn:
Using alpha is identical to using beta at this point. We have not started doing anything like conversion webhooks, the APIs are identical now and the move was to symbolize stability more than anything. In the future this may change but that is the current state at least for 1.6

howardjohn:
So staying on alpha is fine

PR #147, opened 2020-06-16
Move from Istio API v1alpha3 to v1beta1.

JRBANCEL:
Title says it all: addresses #50

As far as I understand, `v1beta1` is 100% the same shape as `v1alpha3`, so the transition should be transparent.


JRBANCEL:
/hold
It seems a bit more complicated than anticipated.
1) The E2E setup doesn't install `v1beta1` CRD
2) We need to be supporting both versions at the same time

JRBANCEL:
This is ready since `v1beta1` is now installed.
/hold cancel

PR #152, opened 2020-06-18
Exclude the webhook from the mesh.

### LINKED ISSUE #150: The webhook must be excluded from the mesh
Currently, if Istio sidecar injection is enabled on `knative-serving` namespace, the webhook will also have a sidecar.
Given that the `MutatingWebhookConfiguration` specifies an expected CA for the certificate exposed by the webhook endpoint, a component outside of the mesh (e.g. `kube-api`) fails to contact the webhook when the sidecar tries to terminate the TLS connection, which happens in mTLS strict mode (in permissive mode, Envoy acts a TCP proxy, see https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/listener.go#L211):
- sidecar + Webhook + mTLS permissive = OK
- sidecar + Webhook + mTLS strict = FAILS
- just Webhook + mTLS permissive = OK
- just Webhook + mTLS strict = OK

The webhook is designed to only be contacted by `kube-api` (outside of the mesh) and to not contact anything, therefore, it should always be excluded from the mesh:
```
kubectl patch deployments.apps -n knative-serving webhook -p '{"spec":{"template":{"metadata":{"annotations":{"sidecar.istio.io/inject":"false"}}}}}'
```

/cc @Cynocracy

JRBANCEL:
Fixes #150 

Right now, testing doesn't cover this, but it will be covered once we move to Istio 1.5 and enable mTLS Strict.
Context: https://github.com/knative/serving/pull/5648

/assign @nak3

### DIFF
--- a/config/webhook-deployment.yaml
+++ b/config/webhook-deployment.yaml
@@ -29,6 +29,9 @@ spec:
     metadata:
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
+        # This must be outside of the mesh to be reachable by
+        # kube-api (also outside the mesh).
+        sidecar.istio.io/inject: "false"
       labels:
         app: istio-webhook
         role: istio-webhook


nak3:
/lgtm
/approve
/hold

Let's wait for a bit while if anyone has other opinion on this and wait for start testing Istio1.5&1.6 instead of 1.4.

JRBANCEL:
> /lgtm

> Let's wait for a bit while if anyone has other opinion on this and wait for start testing Istio1.5&1.6 instead of 1.4.

JRBANCEL:
This is probably not the best way to solve this.
I'll propose another PR.

PR #159, opened 2020-06-24
Explicitly set retry attempt to 0 when no retries are configured in ingress

nak3:
This patch changes to set `istiov1alpha3.HTTPRetry.Attempts` to 0 when
no retries are configued or attempts are set to 0.

Istio tries to retry twice before returning the error by
default. Please refer to the following istio doc:

https://istio.io/latest/docs/concepts/traffic-management/#retries
> The default retry behavior for HTTP requests is to retry twice before returning the error.

fix https://github.com/knative/serving/issues/6549

### DIFF
--- a/pkg/reconciler/ingress/resources/virtual_service.go
+++ b/pkg/reconciler/ingress/resources/virtual_service.go
@@ -220,6 +220,11 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath, 
 			route.Retries.PerTryTimeout = types.DurationProto(http.Retries.PerTryTimeout.Duration)
 		}
 	}
+	if http.Retries == nil || http.Retries.Attempts == 0 {
+		route.Retries = &istiov1alpha3.HTTPRetry{
+			Attempts: 0,
+		}
+	}
 	return route
 }
 
--- a/pkg/reconciler/ingress/resources/virtual_service.go
+++ b/pkg/reconciler/ingress/resources/virtual_service.go
@@ -219,8 +219,7 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath, 
 		if http.Retries.PerTryTimeout != nil {
 			route.Retries.PerTryTimeout = types.DurationProto(http.Retries.PerTryTimeout.Duration)
 		}
-	}
-	if http.Retries == nil || http.Retries.Attempts == 0 {
+	} else {
 		route.Retries = &istiov1alpha3.HTTPRetry{
 			Attempts: 0,
 		}
--- a/pkg/reconciler/ingress/resources/virtual_service.go
+++ b/pkg/reconciler/ingress/resources/virtual_service.go
@@ -211,6 +211,7 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath, 
 		Headers: h,
 	}
 
+	route.Retries = &istiov1alpha3.HTTPRetry{}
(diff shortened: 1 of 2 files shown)


nak3:
I also verified Doug's script in https://github.com/knative/serving/issues/6549 outputs only once.

```
+ grep Failing
Failing 404
Failing 503
Failing 501
Failing 504
```

duglin:
@nak3  thanks!!!!

duglin:
@nak3 what the best way for me to test this? I don't see this code being pulled in under server's vendor dir.

pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -220,6 +220,11 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath,
 			route.Retries.PerTryTimeout = types.DurationProto(http.Retries.PerTryTimeout.Duration)
 		}
 	}
+	if http.Retries == nil || http.Retries.Attempts == 0 {
```
vagababov, line 223: this is just `else` from he previous `if`.

nak3: Thank you. Updated.

nak3:
Hi @duglin You can run `ko apply -f config/` in this repo (`net-istio`) to build and deploy networking-istio like knative serving.

pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -219,6 +219,10 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath,
 		if http.Retries.PerTryTimeout != nil {
 			route.Retries.PerTryTimeout = types.DurationProto(http.Retries.PerTryTimeout.Duration)
 		}
+	} else {
+		route.Retries = &istiov1alpha3.HTTPRetry{
+			Attempts: 0,
+		}
```
vagababov, line 225: if you want to codegolf even more you can just do the following:
```
route.Retries = &istiov1alpha3.HTTPRetry{} // Attempts: 0 is automatic
if http.Retries != nil && ... {
  route.Retries.Retryon = 
  route.Retries.Attempts = ...
  if .. {
  }
}
```

pkg/reconciler/ingress/resources/virtual_service_test.go
```diff
@@ -465,7 +465,9 @@ func TestMakeMeshVirtualServiceSpec_CorrectRetries(t *testing.T) {
 					},
 				}}},
 		},
-		expected: nil,
+		expected: &istiov1alpha3.HTTPRetry{
+			Attempts: 0,
+		},
```
vagababov, line 470:
```suggestion
		expected: &istiov1alpha3.HTTPRetry{},
```
Should work. 
Same below

pkg/reconciler/ingress/resources/virtual_service.go
nak3: Sure, looks better. Updated.

pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -211,6 +211,7 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath,
 		Headers: h,
 	}
 
+	route.Retries = &istiov1alpha3.HTTPRetry{}
```
vagababov, line 214: This way you recreate the HTTPRetry object (c.f my comment earlier). But it's fine — this is not data path.

PR #174, opened 2020-07-10
Implement RewriteHost

julz:
Implementation of https://github.com/knative/serving/issues/7748.

Note: it's not really sensible to try probing with a RewriteHost rule since we can't ensure the hash is what we expect, or even present (since it's returned by a different service), so this just skips probing if a RewriteHost rule is used.

Conformance test in serving: https://github.com/knative/serving/pull/8592.

/assign @tcnghia @mattmoor

### DIFF
--- a/pkg/reconciler/ingress/ingress_test.go
+++ b/pkg/reconciler/ingress/ingress_test.go
@@ -248,8 +248,8 @@ func TestReconcile(t *testing.T) {
 			},
 		},
 		WantCreates: []runtime.Object{
-			resources.MakeMeshVirtualService(insertProbe(ing("no-virtualservice-yet", 1234)), gateways),
-			resources.MakeIngressVirtualService(insertProbe(ing("no-virtualservice-yet", 1234)),
+			resources.MakeMeshVirtualService(context.Background(), insertProbe(ing("no-virtualservice-yet", 1234)), gateways),
+			resources.MakeIngressVirtualService(context.Background(), insertProbe(ing("no-virtualservice-yet", 1234)),
 				makeGatewayMap([]string{"knative-testing/knative-test-gateway", "knative-testing/" + networking.KnativeIngressGateway}, nil)),
 		},
 		WantStatusUpdates: []clientgotesting.UpdateActionImpl{{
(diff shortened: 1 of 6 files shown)


julz:
let me know if this would be easier to review with the two commits split in to different PRs

pkg/reconciler/ingress/ingress_test.go
```diff
@@ -248,8 +248,8 @@ func TestReconcile(t *testing.T) {
 			},
 		},
 		WantCreates: []runtime.Object{
-			resources.MakeMeshVirtualService(insertProbe(ing("no-virtualservice-yet", 1234)), gateways),
-			resources.MakeIngressVirtualService(insertProbe(ing("no-virtualservice-yet", 1234)),
+			resources.MakeMeshVirtualService(context.Background(), insertProbe(ing("no-virtualservice-yet", 1234)), gateways),
```
tcnghia, line 251: May be extract & use something like

```
	ctx := config.ToContext(TestContextWithLogger(t), ReconcilerTestConfig())
```

Since `getGatewayURLFromContext()` uses a `config.Istio` object, which `ReconcilerTestConfig()` already provides.

julz: none of these tests actually do anything with the config.Istio (that's used and tested in reosurces/virtual_service[_test].go) so it seems like a bit more noise than it's worth here. Will happily add if you feel strongly though.

pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -204,9 +205,24 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath,
...
+	var rewrite *istiov1alpha3.HTTPRewrite
+	if http.RewriteHost != "" {
+		rewrite = &istiov1alpha3.HTTPRewrite{
+			Authority: http.RewriteHost,
+		}
+
+		weights = []*istiov1alpha3.HTTPRouteDestination{{
+			Weight: 100,
+			Destination: &istiov1alpha3.Destination{
+				Host: gatewayServiceURLFromContext(ctx, visibility),
```
ZhiminXiang, line 217: Could we always route the traffic to cluster local gateway?

It is likely that users will configure external Gateway to do HTTPS, authentication, or authorization, etc. So It is possible that the external Gateway won't accept the redirected request that has different host.
(not a blocker for this PR, I am fine to do it in a separate PR).

pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -204,9 +205,24 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath,
 		}
 	}
 
+	var rewrite *istiov1alpha3.HTTPRewrite
+	if http.RewriteHost != "" {
```
ZhiminXiang, line 209: It would be helpful to keep the original host header in case the application may want to consume it. WDYT?

julz: since this is relatively low-level (the plan is to build a more user-facing DomainMapping CRD or similar on top of it this feature rather than exposing KIngress to users) I was imagining any behaviour like that would happen higher up (i.e. by adding AppendHeaders to the spec). That way the higher-up code can decide if it wants to add such a header and what to call it. Does that sounds ok or do you think it makes more sense to make it an implicit part of the low-level RewriteHost flow?

pkg/reconciler/ingress/resources/virtual_service.go
julz: yeah, sounds like that may be better. Agree let's get this landed and do it as a follow-on, though, if you're ok with that

pkg/reconciler/ingress/resources/virtual_service.go
ZhiminXiang: That sounds good to me.

PR #178, opened 2020-07-17
Exclude Webhooks from mTLS STRICT

### LINKED ISSUE #150: The webhook must be excluded from the mesh

JRBANCEL:
Fixes #150.

/assign @nak3 @tcnghia

### DIFF
--- a/config/400-webhook-peer-authentication.yaml
+++ b/config/400-webhook-peer-authentication.yaml
@@ -1,0 +1,25 @@
+# Allows the Webhooks to be reached by kube-api with or without
+# sidecar injection and with mTLS PERMISSIVE and STRICT. 
+apiVersion: "security.istio.io/v1beta1"
+kind: "PeerAuthentication"
+metadata:
+  name: "webhook"
+  namespace: "knative-serving"
+spec:
+  selector:
+    matchLabels:
+      app: webhook
+  mtls:
+    mode: PERMISSIVE
+---
+apiVersion: "security.istio.io/v1beta1"
+kind: "PeerAuthentication"
+metadata:
+  name: "istio-webhook"
+  namespace: "knative-serving"
+spec:
+  selector:
+    matchLabels:
+      app: istio-webhook
+  mtls:
+    mode: PERMISSIVE


JRBANCEL:
`PeerAuthentication` is only available in `v1beta1`.
Until Prow installs `v1beta1`, this is blocked.

tcnghia:
/hold 

(see https://github.com/knative-sandbox/net-istio/pull/178#issuecomment-661278950)

JRBANCEL:
It should be unblocked with https://github.com/knative-sandbox/net-istio/pull/185 since it installs `v1beta1`.
/retest

JRBANCEL:
It is ready to merged :rocket:  
/hold cancel
/ping @nak3 @tcnghia

nak3:
@JRBANCEL serving repo still uses Istio 1.4 so we need to merge https://github.com/knative/serving/pull/8214 right?
serving depends on [nigthly net-istio](https://github.com/knative/serving/blob/d556cf7f45d3ed9a703e319599d48235ec543c23/third_party/net-istio.yaml).

JRBANCEL:
/hold
Let's wait for knative/serving#8214

JRBANCEL:
Can't wait for `serving` to just use `net-istio` scripts :pray:

config/400-webhook-peer-authentication.yaml
```diff
@@ -0,0 +1,25 @@
...
+    matchLabels:
+      app: webhook
+  mtls:
+    mode: PERMISSIVE
+---
+apiVersion: "security.istio.io/v1beta1"
+kind: "PeerAuthentication"
+metadata:
+  name: "istio-webhook"
+  namespace: "knative-serving"
```
ZhiminXiang, line 19: should this be under "istio-system" or it does not matter?

JRBANCEL: `The selector determines the workloads to apply the ChannelAuthentication on. If not set, the policy will be applied to all workloads in the same namespace as the policy.`

So in this case it doesn't matter since a selector is specified.
Plus it is cleaner to have it in our namespace, since it is about resources in our namespace.

ZhiminXiang: SGTM. thanks!

PR #185, opened 2020-07-21
Use Istio Operator.

### LINKED ISSUE #181: #175 is incorrectly still using 1.4 CRDs
#175 is not doing what it is supposed to do because in Istio 1.6.x, there is no more Helm charts, so `download-istio.sh` doesn't work as expected.

/assign

JRBANCEL:
Fixes #181 

* Move from Istio 1.5.0 to 1.5.7
* Replace the `download-istio.sh` + checked-in YAML + `kubectl apply` with `install-istio.sh` which uses `istioctl install` and profiles.

This is the recommended installation method. Using `istioctl manifest generate` and `kubectl apply -f` is explicit [not recommended](https://istio.io/latest/docs/setup/install/istioctl/#generate-a-manifest-before-installation).

Profiles are directly inspired by @nak3  work in `serving` 🙏 .
How can we avoid duplication? Ideally, `serving` would use `net-istio` installation script but the dependency is the wrong way. EDIT: @tcnghia is working on inverting the dependency, so `serving` will be able to call `install-istio.sh` 😀.


third_party/istio-1.5.7/install-istio.sh
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Download and unpack Istio
+ISTIO_VERSION=1.5.7
+ISTIO_TARBALL=istio-${ISTIO_VERSION}-linux.tar.gz
+DOWNLOAD_URL=https://github.com/istio/istio/releases/download/${ISTIO_VERSION}/${ISTIO_TARBALL}
+
+wget --no-check-certificate $DOWNLOAD_URL
```
nak3, line 22: The `istio-1.6.5-linux-amd64.tar.gz` is around 40M and it takes 5 to 10 min (on my local).
It is alright to download the tar file every single e2e tests?
As you might known, in the serving repo I generated the manifest by `istioctl manifest generate` to avoid it.

JRBANCEL: Fair remark for slow internet connections and local development.

You can see in the logs that in the CI system takes ~1s to download the package.
My concern with `istioctl generate manifest` + `kubectl apply` is that unlike `istioctl install` it doesn't have ordering guarantees and is actually not recommended. See https://istio.io/latest/docs/setup/install/istioctl/#generate-a-manifest-before-installation.

> The output from manifest generate can also be used to install Istio using kubectl apply or equivalent. However, these alternative installation methods may not apply the resources with the same sequencing of dependencies as istioctl install and are not tested in an Istio release.

JRBANCEL: But 5 to 10 minutes is insane. That's like 60 kBps 😢. Is there an issue with Github and Japan somehow?

nak3: I see. Yeah, then there are some issue in Github or Japan :smile:

PR #190, opened 2020-07-23
Probe a single host instead of every hosts

### LINKED ISSUE #188: Probing optimization take 2
_This is the net-istio side of https://github.com/knative/serving/issues/8765._

Currently the network probing logic implements an optimization where a pod is assumed to have been fully programmed once any part of the network programming we are probing has been successfully verified.  In practice, this means that for each pod we are enqueuing N pieces of work and cancelling N-1 pieces of work for each gateway pod every time we change a kingress.

Given that this builds on a faulty assumption (in the linked issue), I'd propose instead optimizing this on the net-istio side here:

https://github.com/knative-sandbox/net-istio/blob/56f23ab0d69625e409602c9bd56c9de07f008142/pkg/reconciler/ingress/lister.go#L105

Instead of enqueuing N pieces of work per pod, and then cancelling the N-1 pieces of work, simply choose one piece of work to verify (per IP) and prune the rest.  This should achieve what the prober-side optimization yields and more because this should also improve the throughput of the prober workqueue.

I'd like to avoid having Istio take a hit to correct the issue above, and this should be safe to start in isolation, so if anyone wants to try and knock this off I'd appreciate it. 🙏 

cc @JRBANCEL @tcnghia @vagababov

JRBANCEL:
Fixes #188.

# Changes
* Renamed `gatewayHosts` to `hostsByGateway` because it is clearer
* Pick the first host instead of all of them

### DIFF
--- a/pkg/reconciler/ingress/lister.go
+++ b/pkg/reconciler/ingress/lister.go
@@ -67,11 +67,12 @@ func (l *gatewayPodTargetLister) ListProbeTargets(ctx context.Context, ing *v1al
 	}
 
 	results := []status.ProbeTarget{}
-	gatewayHosts := ingress.HostsPerVisibility(ing, qualifiedGatewayNamesFromContext(ctx))
-	gatewayNames := []string{}
-	for gatewayName := range gatewayHosts {
+	hostsByGateway := ingress.HostsPerVisibility(ing, qualifiedGatewayNamesFromContext(ctx))
+	gatewayNames := make([]string, 0, len(hostsByGateway))
+	for gatewayName := range hostsByGateway {
 		gatewayNames = append(gatewayNames, gatewayName)
 	}
+
 	// Sort the gateway names for a consistent ordering.
 	sort.Strings(gatewayNames)
 	for _, gatewayName := range gatewayNames {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/ingress/lister.go
```diff
@@ -91,14 +92,14 @@ func (l *gatewayPodTargetLister) ListProbeTargets(ctx context.Context, ing *v1al
...
-			// Use sorted hosts list for consistent ordering.
-			for i, host := range gatewayHosts[gatewayName].List() {
-				newURL := *target.URLs[0]
-				newURL.Host = host + ":" + target.Port
-				qualifiedTarget.URLs[i] = &newURL
+				URLs:    make([]*url.URL, 1),
 			}
+			// Pick a single host since they all end up being used in the same
+			// VirtualService and will be applied atomically by Istio.
+			host := hostsByGateway[gatewayName].List()[0]
```
yanweiguo, line 99: Do we need to make sure the list is not empty?

JRBANCEL: Good question. I should maybe add a comment about this.
`ingress.HostsPerVisibility` guarantees it.

PR #199, opened 2020-07-30
Stop using knative.dev/serving in cmd/ and pkg/

tcnghia:
This should leave conformance testing as the only knative.dev/serving dependency left. I will follow up with another PR.


pkg/reconciler/ingress/resources/util.go
```diff
@@ -32,6 +32,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	ServingGroupName       = "serving.knative.dev"
```
vagababov, line 36: You have this string declared twice?

pkg/defaults/deployment_defaults.go
```diff
@@ -23,8 +23,23 @@ import (
...
 	"knative.dev/pkg/apis"
+)
+
+const (
+	// ServingGroupName is the group name for knative labels and annotations
+	ServingGroupName = "serving.knative.dev"
+
+	// ServingServiceLabelKey is the label key attached to a Route and Configuration indicating by
+	// which Service they are created.
+	ServingServiceLabelKey = ServingGroupName + "/service"
```
vagababov, line 34: Those don't seem to be used crosspackage, so probably lowecase?

vagababov:
LG, but tests look unhappy

pkg/defaults/deployment_defaults.go
tcnghia: Done

pkg/reconciler/ingress/resources/util.go
tcnghia: Yeah, the defaults package and the reconciler/ packages are quite independent so I didn't want to refer to one from the other.

PR #209, opened 2020-08-06
Change to install-istio.sh to use commonly

nak3:
This patch updates install-istio.sh to support to adjust version and manifest path (URL/file path).

e.g.
```
$ ISTIO_VERSION=1.5.7
$ ./third_party/install-istio.sh third_party/istio-stable/istio-ci-no-mesh.yaml

$ ISTIO_VERSION=1.6.5
$ ./third_party/install-istio.sh https://raw.githubusercontent.com/knative-sandbox/net-istio/master/third_party/istio-latest/istio-ci-no-mesh.yaml

$ ISTIO_VERSION=stable
$ ./third_party/install-istio.sh https://raw.githubusercontent.com/knative-sandbox/net-istio/master/third_party/istio-stable/istio-ci-no-mesh.yaml
```

The goal is to use `install-istio.sh` from serving repo. It will be able to share istio manifest between net-istio and serving as reported in https://github.com/knative/serving/issues/8896.


nak3:
/hold

Ugh... This PR might disturb @JRBANCEL 's https://github.com/knative/serving/pull/8945 as I changed `third_party/istio-${ISTIO_VERSION}/install-istio.sh` to `third_party/install-istio.sh`.
@JRBANCEL what do you think about this change? You don't need it?

JRBANCEL:
Using the path or an env variable is pretty similar. The path is slightly more explicit.
Up to you but I would just not make the change since my PR moving to channels will get rid of versions altogether.

nak3:
Thank you, I understand.
Then, let's not merge it for now :smile:

PR #222, opened 2020-08-13
Run 5 parallel invocations for KIngress conformance tests.

tcnghia:
This will bring the load a bit  closer to the level of concurrency we see in Serving e2e tests.

### DIFF
--- a/test/conformance/ingress_test.go
+++ b/test/conformance/ingress_test.go
@@ -19,11 +19,19 @@ limitations under the License.
 package conformance
 
 import (
+	"strconv"
 	"testing"
 
 	"knative.dev/networking/test/conformance/ingress"
 )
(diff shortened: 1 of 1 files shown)


mattmoor:
You should see how long the `go test` invocation takes at 5.  To go to 11 I needed to bump the `-timeout` in net-contour.

You may want to `/test` this a couple times before it lands.

/lgtm

vagababov:
Interesting. Can we vary a parameter to make it more interesting? :)

vagababov:
Also can make it a flag, so it's easier to change, but 🤷

tcnghia:
23-24 minutes for 5x. I'd let this go in first and watch for a couple runs.

/unhold

PR #223, opened 2020-08-14
Turn it up to 11

mattmoor:
😈

/hold

PR #226, opened 2020-08-15
Enable alpha/beta tests

mattmoor:


mattmoor:
fixed the argument order 🙄

julz:
Fun :(. Think we might want to bump knative.dev/networking to pick up https://github.com/knative/networking/pull/90 since these tests are [turned up to 11](https://github.com/knative-sandbox/net-istio/pull/223) .

julz:
.. and https://github.com/knative/networking/pull/96, scheme got dropped in knative/networking#90 :).

mattmoor:
We can cherrypick that and your PR, but I suspect that 503 may be lack of probing.

julz:
Possibly, tho I got same errors locally before manually bumping vendor, and then ran three times in a row without (not that lack of probing may not also be a problem).

mattmoor:
👍 Yeah Istio may fail differently than Contour.  When Contour sees conflicts neither get sent to Envoy and you end up with 404s.

julz:
yeh looks like it, I get reliable 503s without the bump once `iterations > 1` (and passed normally 20 times in a row after bumping vendor) - guess istio gets confused later in the process than contour does :)

Issue #212, opened 2020-08-10
Remove cluster-local-gateway Deployment


JRBANCEL:
The details are in this [doc](https://docs.google.com/document/d/1kqGPwEeo2uNfZlVlS3wZNeCiGakRxTXuclz9rehw9j8/).

This is to track the rollout. The current plan to have 0 downtime is:

## Phase 1
1) Add a `server` entry to `cluster-local-gateway` on port 8081 (no-op just makes the Envoy listen on this port)

🕐  Wait for Istio propagation

## Phase 2
2) Modify the `cluster-local` `Service` `targetPort` to `8081` (k8s will route internal traffic to the new port) 
3) Remove the original `server` entry in `cluster-local-gateway` (port 80 is not used anymore)
4) Update the selector of the `Gateway` to select both `Deployment` (Ingress and Cluster Local)

🕐 : Wait for Istio propagation

## Phase 3
5) Modify the `cluster-local` `Service` to select the Ingress `Deployment`
5) Update the selector of the `Gateway` to select only the Ingress `Deployment`

:clock1: Wait for Istio propagation

## Phase 4
6) Delete Cluster Local `Deployment` 

This could be done in a single release by waiting for propagation before proceeding to the next steps where required, or we can spread it over multiple releases. @tcnghia any thoughts?

/assign

tcnghia:
My inclination is to avoid staging over multiple releases: we can get result faster and even allow N+2 upgrades without issue.

JRBANCEL:
Ok, let me see how much of a downtime it would be we do all the changes at the same time.
It should be minimal and less error prone. Also, it is for internal traffic only so less visible for external calls.

JRBANCEL:
I am writing a migration job. It is simpler and safer.

JRBANCEL:
Phase 3 is done: #411 
/close

PR #237, opened 2020-08-21
Add new local Gateway sharing same deployment as ingress Gateway.

JRBANCEL:
Related to https://github.com/knative-sandbox/net-istio/issues/212.

The idea is to create a new `Gateway` named `knative-local-gateway` (more consistent with `knative-ingress-gateway` than `cluster-local-gateway`). This Gateway binds to the same Envoy Pods as the Istio ingress Gateway, but on a different port (`8081` because `8080` is already used by Istio). There is a corresponding Kubernetes Service mapping named `knative-local-gateway` living in `knative-serving` (cleaner than `cluster-local-gateway` living in `istio-system`) mapping port `80` to port `8081`.
`knative-local-gateway` is added to the default Istio Gateways, this way the VirtualServices of local Knative Services bind to it (i.e. ready to serve traffic), but because the `KIngress` status uses the first local Gateway in its status, the Services of type `ExternalName` still point to `cluster-local-gateway` (and it is not accessible to external requests since `8081` is not exposed through the `LoadBalancer` Service).

Effectively, this is a no-op, but in the next release we can remove reference to `cluster-local-gateway` in default `config-istio` and there will be no traffic interruption and the requests will flow to `knative-local-gateway`.

After that, everything related to `cluster-local-gateway` can be safely removed, including the `Deployment`, which is the end goal here.

/hold
/assign @tcnghia 
/assign @ZhiminXiang 
Let me know what you think.

### DIFF
--- a/pkg/reconciler/ingress/config/istio.go
+++ b/pkg/reconciler/ingress/config/istio.go
@@ -41,8 +41,11 @@ const (
 	// KnativeIngressGateway is the name of the ingress gateway
 	KnativeIngressGateway = "knative-ingress-gateway"
 
-	// ClusterLocalGateway is the name of the local gateway
+	// ClusterLocalGateway is the name of the local gateway (will be deprecated)
 	ClusterLocalGateway = "cluster-local-gateway"
+
+	// KnativeLocalGateway is the name of the local gateway (will be promoted)
+	KnativeLocalGateway = "knative-local-gateway"
 )
 
 func defaultIngressGateways() []Gateway {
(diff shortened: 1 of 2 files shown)


JRBANCEL:
> But we may still want to double check that users cannot access cluster-local ingress from outside of the cluster just in case.

Yes, I tested that. Since the cluster local Gateway is on a port that is not exposed internally, it can't be accessed.
Either you hit `<PUBLIC_IP>:8081` and you can't connect, or you hit `<PUBLIC_IP>:80` and it goes to the public Gateway and you get a 404 for a cluster local service.

tcnghia:
@JRBANCEL is this ready for `/unhold`?

tcnghia:
Also I assigned you https://github.com/knative/serving/issues/6269 since I thought it would be related to this work.

pkg/reconciler/ingress/config/istio.go
```diff
@@ -60,6 +63,11 @@ func defaultLocalGateways() []Gateway {
 		Name:      ClusterLocalGateway,
 		ServiceURL: fmt.Sprintf(ClusterLocalGateway+".istio-system.svc.%s",
 			network.GetClusterDomainName()),
+	}, {
+		Namespace: system.Namespace(),
+		Name:      KnativeLocalGateway,
+		ServiceURL: fmt.Sprintf(KnativeLocalGateway+".knative-serving.svc.%s",
```
mattmoor, line 69: This should be using `system.Namespace()`

mattmoor: In fact, we have a helper for synthesizing service hostnames that this should be using too in place of `GetClusterDomainName()`

Issue #259, opened 2020-09-04
ingress pods cannot be started with third_party/istio-1.5.7-helm/istio-minimal.yaml


ZhiminXiang:
I have tried to apply the third_party/istio-1.5.7-helm/istio-minimal.yaml. But both istio-ingressgateway pods and cluster-local-gateway pods cannot be started because istio-proxy container is not ready.

```
cluster-local-gateway-96488b8bf-qknql   0/1     Running   0          98s
istio-ingressgateway-6c866f94c6-vhpxv   1/2     Running   0          98s
istio-pilot-6bdfc6f49c-k2hfd            1/1     Running   0          98s
```

I can reproduce this issue consistently.

I checked the log, and got a lot of the following logs:
```
2020-09-04T03:45:21.061808Z	info	waiting for file
2020-09-04T03:45:21.162125Z	info	waiting for file
2020-09-04T03:45:21.262440Z	info	waiting for file
2020-09-04T03:45:21.362749Z	info	waiting for file
2020-09-04T03:45:21.462978Z	info	waiting for file
2020-09-04T03:45:21.563242Z	info	waiting for file
2020-09-04T03:45:21.663541Z	info	waiting for file
```

And later I got the following logs:
```
2020-09-04T04:10:25.459577Z	info	Envoy proxy is NOT ready: config not received from Pilot (is Pilot running?): cds updates: 0 successful, 1 rejected; lds updates: 1 successful, 0 rejected
```

This issue caused https://github.com/knative/serving/issues/9275

/cc @nak3

ZhiminXiang:
I also noticed the error message from the istio-proxy container:
```
2020-09-04T05:47:43.333146Z	info	waiting 2m0s for /etc/certs/key.pem
2020-09-04T05:49:43.376058Z	warn	file still not available after2m0s
```
I suspect this issue is caused by missing secrets:
1. istio.istio-ingressgateway-service-account for istio ingress
2. istio.cluster-local-gateway-service-account for cluster local gateway

nak3:
@ZhiminXiang 
it is same issue with https://github.com/knative/serving/issues/6254.
Since istio-minimal does not set `global.mtls.auto=false` explicitly, cluster-local-gateway tries to find a cert. istio-lean also does not have it, but it deploys citadel so no issue.
I will send the PR.

PR #260, opened 2020-09-04
Add global.mtls.auto false explicitly in values-local.yaml and values-lean.yaml

nak3:
This patch add global.mtls.auto to false explicitly in
values-local.yaml and values-lean.yaml.

Without `global.mtls.auto=false`, helm sets
`ISTIO_AUTO_MTLS_ENABLED=true` and gateway tries to find certificates.
It can be solved by specify `global.mtls.auto=false` or deploying
citadel.

/cc @ZhiminXiang 

fix https://github.com/knative-sandbox/net-istio/issues/259


tcnghia (approved):
/lgtm

tcnghia:
Hmm, according to the links the tests are all passing but for some reasons the status not reflected back.

PR #270, opened 2020-09-09
Fix the incorrect unit test when auto TLS is enabled

ZhiminXiang:
We need to set visibility explicitly in the unit test so that the gateway corresponding to the visibility could be picked up

/assign @JRBANCEL


pkg/reconciler/ingress/ingress_test.go
```diff
@@ -163,6 +164,7 @@ var (
 				}},
 			},
 		},
+		Visibility: v1alpha1.IngressVisibilityExternalIP,
```
nak3, line 167: :+1:  
we missed this!

PR #287, opened 2020-09-16
[release-0.17] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign tcnghia
/cc tcnghia


tcnghia:
/hold

I am not sure if we want to cause a release just because of a vendor change in `test-infra` 
cc @nak3 @ZhiminXiang

ZhiminXiang:
If this is the only for `test-infra` update, I think we should close this one.

vagababov:
I have a more interesting question: why did we merge test-infra into release branch in the first place?

/cc @chizhg @chaodaiG

chizhg:
> I have a more interesting question: why did we merge test-infra into release branch in the first place?
> 
> /cc @chizhg @chaodaiG

We started cutting release for test-infra repo since release-0.13 by following the same schedule as all other Knative repos, IIRC it was because ppl had concerns on the backward compatibility for the scripts under test-infra repo.

PR #288, opened 2020-09-17
Include HEAD commit hash in release YAML.

JRBANCEL:
First part of https://github.com/knative/serving/issues/9420.

This way we can clone `net-istio` in `serving` at the exact point `net-istio.yaml` was generated.
No more hard-coding in the scripts.

/assign @tcnghia


tcnghia (approved):
/lgtm

PR #301, opened 2020-09-21
Check that KIngress is ready before probing

### LINKED ISSUE #262: net-istio should guard the prober against resync flooding
Right now the prober is unconditionally called during reconciliation to determine readiness: https://github.com/knative-sandbox/net-istio/blob/81a9b95df6ba5516bb3240b53caaa1686fc2dd75/pkg/reconciler/ingress/ingress.go#L203

This is ok under normal circumstances because internally the prober caches the probe result and so on global resyncs we hit the cache and things finish quickly.  However, when we resync due to failing over (perhaps due to rollout) this cache is empty.

If I have 1000 ksvc, they are exposed on both gateways, and those gateways have 5 pods each, then the prober must perform `5 * 2 * 1,000 = 10,000` probes before normal work may resume.

In `net-contour`, I changed things to have the main Reconcile loop rely on the recorded readiness of the `kingress` to elide `IsReady` checks, effectively recording the probers internal cache in the CRD's durable status.

cc @JRBANCEL @tcnghia @nak3 @ZhiminXiang

arturenault:
Fixes #262

- Implementation and comments copied from net-countour.

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -211,7 +211,7 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 		// As this is an optimization, we don't worry about the ObservedGeneration
 		// skew we might see when the resource is actually in flux, we simply care
 		// about the steady state.
-		logger.Debug("kingress is ready, skipping probe.")
+		logger.Debug("Kingress is ready, skipping probe.")
 		ready = true
 	} else {
 		readyStatus, err := r.statusManager.IsReady(ctx, ing)
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -109,7 +109,6 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	ing.Status.InitializeConditions()
 	logger.Infof("Reconciling ingress: %#v", ing)
 
-	ing.Status.ObservedGeneration = ing.GetGeneration()
 	gatewayNames := qualifiedGatewayNamesFromContext(ctx)
 	if r.shouldReconcileTLS(ctx, ing) {
 		originSecrets, err := resources.GetSecrets(ing, r.secretLister)
@@ -229,6 +228,8 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	} else {
 		ing.Status.MarkLoadBalancerNotReady()
 	}
+
+	ing.Status.ObservedGeneration = ing.GetGeneration()
 
 	// TODO(zhiminx): Mark Route status to indicate that Gateway is configured.
 	logger.Info("Ingress successfully synced")
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -201,16 +201,29 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	// Update status
 	ing.Status.MarkNetworkConfigured()
 
-	ready, err := r.statusManager.IsReady(ctx, ing)
-	if err != nil {
-		return fmt.Errorf("failed to probe Ingress %s/%s: %w", ing.GetNamespace(), ing.GetName(), err)
-	}
-	if ready {
-		publicLbs := getLBStatus(publicGatewayServiceURLFromContext(ctx))
-		privateLbs := getLBStatus(privateGatewayServiceURLFromContext(ctx))
-		ing.Status.MarkLoadBalancerReady(publicLbs, privateLbs)
+	if ing.IsReady() {
+		// When the kingress has already been marked Ready for this generation,
+		// then it must have been successfully probed.  The status manager has
+		// caching built-in, which makes this exception unnecessary for the case
+		// of global resyncs.  HOWEVER, that caching doesn't help at all for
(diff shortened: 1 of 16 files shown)


tcnghia:
@arturenault thanks for fixing! can we please add a test for this new case?

arturenault:
> @arturenault thanks for fixing! can we please add a test for this new case?

Yup will do. Still debugging conformance test failures as well

arturenault:
Having some trouble writing a unit test here. Basically the idea is to verify whether statusManager.IsReady is called, right?

It looks like the TableTest framework isn't directly for this as it's focused on monitoring resources that k8s knows about. I could use it if I injected a fake StatusManager into the context, stored information in it about the calls made to it, and then verify in a PostCondition that the number of calls made matches what we expect. This would be similar to what's done [here](https://github.com/knative/serving/blob/9d0250fcaaecc6b4c0370dd7eaef4ae73d6c8624/pkg/reconciler/route/table_test.go#L1233-L1235), but it would require moving the fakeStatusManager into a more public location, probably into knative/networking/pkg/client/injection/client.

That or I could write a custom test that does these sorts of tests. Either is fine; but both sound like a lot of work for a relatively simple optimization, especially given that we might be getting rid of probers soon with istio VirtualService's Status field.

@tcnghia @JRBANCEL @ZhiminXiang thoughts?

arturenault:
> Having some trouble writing a unit test here. Basically the idea is to verify whether statusManager.IsReady is called, right?
> 
> It looks like the TableTest framework isn't directly for this as it's focused on monitoring resources that k8s knows about. I could use it if I injected a fake StatusManager into the context, stored information in it about the calls made to it, and then verify in a PostCondition that the number of calls made matches what we expect. This would be similar to what's done [here](https://github.com/knative/serving/blob/9d0250fcaaecc6b4c0370dd7eaef4ae73d6c8624/pkg/reconciler/route/table_test.go#L1233-L1235), but it would require moving the fakeStatusManager into a more public location, probably into knative/networking/pkg/client/injection/client.
> 
> That or I could write a custom test that does these sorts of tests. Either is fine; but both sound like a lot of work for a relatively simple optimization, especially given that we might be getting rid of probers soon with istio VirtualService's Status field.
> 
> @tcnghia @JRBANCEL @ZhiminXiang thoughts?

On further thought I think I'll go aheead with moving the fakeStatusManager. Having more robust mocking/injection will help with testing this and other networking backends, and will help with the work on VirtualService Status.

vendor/knative.dev/networking/pkg/status/status.go
```diff
@@ -151,7 +151,8 @@ func NewProber(
 	}
 }
 
-func ingressKey(ing *v1alpha1.Ingress) string {
+// IngressKey returns a key that uniquely identifies an Ingress object
```
JRBANCEL, line 154: Why change the visibility of this function?

JRBANCEL:
I see a bunch of cosmetic changes. These changes should be in their own PR, it makes review simpler.

arturenault:
> I see a bunch of cosmetic changes. These changes should be in their own PR, it makes review simpler.

Yes sorry, this is not ready for review yet. The changes you're seeing are mostly thing being pulled in from knative/networking to unblock me since master is currently pinned.

arturenault:
not sure why this still says needs-rebase and do-not-merge/draft... I've rebased and set ready to review

pkg/reconciler/ingress/ingress.go
```diff
@@ -213,6 +229,8 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 		ing.Status.MarkLoadBalancerNotReady()
 	}
 
+	ing.Status.ObservedGeneration = ing.GetGeneration()
```
JRBANCEL, line 232: Maybe document why this needs to be here.
Also, as suggested it might be safer to copy the ingress object so it is crystal clear when we mutate the new one and query the old one.

arturenault: I think I prefer this approach; it seems unnecessary to perform an extra copy to get the same result. Plus it'd add confusion as to which version we need to be processing on.

IMO the ObservedGeneration update should always have been here, as we weren't following the documentation for ObservedGeneration's purpose:
```
	// ObservedGeneration is the 'Generation' of the Service that
	// was last processed by the controller.
	// +optional
```
Before, we were changing it before processing. So if anything happened during processing we would've updated it incorrectly. I'll add a comment though

pkg/reconciler/ingress/ingress.go
```diff
@@ -201,10 +200,27 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
...
+		// When the kingress has already been marked Ready for this generation,
+		// then it must have been successfully probed.  The status manager has
+		// caching built-in, which makes this exception unnecessary for the case
+		// of global resyncs.  HOWEVER, that caching doesn't help at all for
+		// the failover case (cold caches), and the initial sync turns into a
+		// thundering herd.
+		// As this is an optimization, we don't worry about the ObservedGeneration
+		// skew we might see when the resource is actually in flux, we simply care
+		// about the steady state.
+		logger.Debug("kingress is ready, skipping probe.")
```
vagababov, line 214:
```suggestion
		logger.Debug("Kingress is ready, skipping probe.")
```

vagababov:
This is lgtm from me. But I'll leave to the real experts :)

arturenault:
@vagababov @JRBANCEL looks like this is still missing LGTM label (approve is good). can one of you LGTM?

PR #293, opened 2020-09-17
Add NodePort Service profiles.

JRBANCEL:
This allows to deploy Istio with `istio-ingressgateway` Service type `NodePort` instead of `LoadBalancer`.
This is needed for `Kind` and `minikube`.

Fixes #285.

/assign @tcnghia

PR #303, opened 2020-09-22
Rename and tweak Kind profiles.

JRBANCEL:
In #293, I introduced profiles exposing the Ingress Service has a Node Port so that it would work with the Kind E2E running in a GitHub Action. GitHub Actions have only 2 cores, so the `resource.requests` and `replicaCount` need to be scaled way down in order for Istio to be deployed successfully on Kind (https://github.com/knative/serving/pull/9471).

Also, this renames the profiles to make the intent clearer.


tcnghia (approved):
/lgtm

PR #306, opened 2020-09-22
Tune CodeCov down a little bit.

tcnghia:
Currently CodeCov is too strict causing harmless PRs (update deps) to be flagged. This mirrors what we have in Serving and tune things down a little bit.

/assign @mattmoor


mattmoor (commented):
/lgtm

PR #313, opened 2020-09-25
Move the stable back to 1.4

ZhiminXiang:
The change made in stable is based on the commit f64ed34d3776a444372483dddc15a330c6c1ac53 https://github.com/knative-sandbox/net-istio/tree/f64ed34d3776a444372483dddc15a330c6c1ac53/third_party/istio-1.4.6

The motivation to move stable back to 1.4 is we can import this commit to serving so that serving E2E tests could test {1.4, 1.7} instead of {1.4, 1.5}.

Per discussion: https://github.com/knative/serving/pull/9471

/assign @JRBANCEL 
/assign @nak3


JRBANCEL:
This removes the `istio-kind-{no-}mesh.yaml` profiles used with Kind and GitHub Action :-\

JRBANCEL:
I can re-add later, it is not required until the GitHub Action PR is merged.

JRBANCEL:
Also, do we need to this?
As long as we don't move the pinned version in `serving`, it will (and is) still be using 1.4.x.

ZhiminXiang:
I think the motivation here is we try to promote Istio latest to 1.7 to replace the current latest 1.5 for serving E2E test in the release 0.18.

tcnghia:
@ZhiminXiang @JRBANCEL now that we cut the release, may be this is no longer needed and we will test 1.7 as stable/latest?

ZhiminXiang:
we can drop this PR and try to bump to 1.7.

PR #317, opened 2020-09-29
Fix the format here

vagababov:
As a departure from the standard procedure this one actually was broken
and required manual interaction :-)

/assign @tcnghia @ZhiminXiang


tcnghia (approved):
/lgtm

PR #320, opened 2020-09-30
Specify port for rewrite host destination

julz:
When the incoming request is using https, failing to specify the destination port means we end up re-running the request against port 443 on the cluster-local-gateway which is not what we want.  We see errors like "`Connection #0 to host blah-blah left intact
upstream connect error or disconnect/reset before headers. reset reason: connection failure`" when using a rewrite rule over https.  

Explicitly specifying port 80 for the destination fixes our issue (I think the hard-coding of 80 is fine here because this is the only port the cluster local gateway can possibly be on).

This matches the way we already set a port for regular non-rewrite-host splits (see [virtual_service.go#L189](https://github.com/knative-sandbox/net-istio/blob/9fb6d8eee0f12b6cc3bfbc34b646246ef45eafd4/pkg/reconciler/ingress/resources/virtual_service.go#L189)).

/assign @tcnghia @nak3

### DIFF
--- a/pkg/reconciler/ingress/resources/virtual_service.go
+++ b/pkg/reconciler/ingress/resources/virtual_service.go
@@ -215,7 +215,7 @@ func makeVirtualServiceRoute(ctx context.Context, hosts sets.String, http *v1alp
 			Destination: &istiov1alpha3.Destination{
 				Host: privateGatewayServiceURLFromContext(ctx),
 				Port: &istiov1alpha3.PortSelector{
-					Number: 80,
+					Number: GatewayHTTPPort,
 				},
 			},
 		}}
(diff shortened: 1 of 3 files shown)


pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -214,6 +214,9 @@ func makeVirtualServiceRoute(ctx context.Context, hosts sets.String, http *v1alp
 			Weight: 100,
 			Destination: &istiov1alpha3.Destination{
 				Host: privateGatewayServiceURLFromContext(ctx),
+				Port: &istiov1alpha3.PortSelector{
+					Number: 80,
```
tcnghia, line 218: Is there a constant we can use?

tcnghia (changes requested):

pkg/reconciler/ingress/resources/virtual_service.go
julz: I dont think so unfortunately (but Id be happy to use one if you point can me at one!), I can create one but it'd be something like `const clusterGatewayHTTPPort = 80` so maybe not adding much?

tcnghia: May be extract pkg/reconciler/ingress/resources/gateway.go#L314 and use the same?

julz: 👍️ done.

tcnghia (approved):
/lgtm

PR #352, opened 2020-10-12
Remove unused 'generation' parameter from ingress_test

arturenault:


pkg/reconciler/ingress/ingress_test.go
```diff
@@ -1167,7 +1167,7 @@ func ReconcilerTestConfig() *config.Config {
 	}
 }
 
-func ingressWithStatus(name string, generation int64, status v1alpha1.IngressStatus) *v1alpha1.Ingress {
```
arturenault, line 1170: this is the base function used by this test to create ingresses. generation is all over this file because it's a parameter here, but it's not used at all.

vagababov:
/ok-to-test

wtf

nak3 (approved):

PR #363, opened 2020-10-17
Add the instruction to install knative serving and net-istio

ZhiminXiang:


DEVELOPMENT.md
```diff
@@ -76,3 +76,20 @@ Run the following command to install Istio for development purpose:
 ```shell
 third_party/istio-stable/install-istio.sh istio-ci-no-mesh.yaml
 ```
+
+### Install Knative Serving
+
+Run the following command to install Knative
+
+```shell
+kubectl apply -f https://github.com/knative/serving/releases/download/v0.18.0/serving-crds.yaml
```
nak3, line 85: This is a development guide, so would it be better to use nightly build like:

```
kubectl apply --filename https://storage.googleapis.com/knative-nightly/serving/latest/serving-crds.yaml
kubectl apply --filename https://storage.googleapis.com/knative-nightly/serving/latest/serving-core.yaml
```

Also, the version (`0.18.0`) needs to be updated very often so I am prefer to use the nightly build manifest.

ZhiminXiang: SGTM. Done

nak3 (approved):
/lgtm

lint error is not related to this PR. I will create another PR for it.

PR #365, opened 2020-10-19
Remove Ingress Timeout

nak3:
Ingress's timeout was set to super long timeout by
https://github.com/knative/serving/pull/8965.

But the current conformance test for [omitted timeout ](https://github.com/knative/networking/pull/157) proved that we can drop
the timeout so knative/serving and knative/networking are dropping it.
Please see https://github.com/knative/networking/pull/228 and https://github.com/knative/serving/pull/9852.

Hence this patch removes timeout in net-istio to follow up them.


tcnghia (approved):
/lgtm

Issue #344, opened 2020-10-09
Run net-istio tests on KinD


tcnghia:
That allows us to test more K8s version and customizations like cluster sufffix (see https://github.com/knative-sandbox/net-istio/issues/343 )

nak3:
This was solved by https://github.com/knative-sandbox/net-istio/pull/394 :tada:

Issue #343, opened 2020-10-09
Test `net-istio` with a cluster with non-default cluster suffix


tcnghia:
In https://github.com/knative/networking/issues/209 we allow the cluster suffix to be specified.

Once `net-istio` tests are on `KinD` we can use a similar workflow to that of https://github.com/knative-sandbox/net-contour/actions/runs/296836515/workflow to test net-istio with a non-default cluster suffix.

PR #383, opened 2020-10-28
Add clusterDomain to values.global.proxy in kind yaml to replace it easily

nak3:
When k8s cluster is changed the cluster.local domain, Istio (istio gateway
and sidecar) needs to change the domain by values.global.proxy.

This patch adds the value to replace it easily during KinD e2e.

Please see also https://github.com/knative/serving/issues/9750

/cc @mattmoor @tcnghia @ZhiminXiang @JRBANCEL

### DIFF
--- a/third_party/istio-latest/istio-kind-mesh.yaml
+++ b/third_party/istio-latest/istio-kind-mesh.yaml
@@ -19,6 +19,7 @@ spec:
     global:
       proxy:
         autoInject: enabled
+        clusterDomain: cluster.local
       useMCP: false
       # The third-party-jwt is not enabled on all k8s.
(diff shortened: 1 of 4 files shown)


nak3:
The value will be replaced by the script as https://github.com/knative/serving/pull/9969.

mattmoor:
@nak3 We should set up KinD testing on this repo, so that we can more easily test these things closer to the source.  Thanks for the quick fix, as always 🤩 

/lgtm

tcnghia:
Thanks a lot @nak3

tcnghia:
@mattmoor @nak3  I filled the feature request here https://github.com/knative-sandbox/net-istio/issues/344 and https://github.com/knative-sandbox/net-istio/issues/343

PR #388, opened 2020-10-28
Move knative-local-gateway service to istio-system.

JRBANCEL:
There was a fatal in [my plan](https://github.com/knative-sandbox/net-istio/pull/237#issue-471304463) to collocate both Ingress and Local Gateways into the same Envoy Pods: k8s service selectors only work inside the same namespace (the Envoy Pods are in `istio-system`).

This moves the `knative-local-gateway` k8s service to `istio-system`, this is fine because this service is not being used yet (see phase #2 of the plan).

Also, this addresses @mattmoor [comment](https://github.com/knative-sandbox/net-istio/pull/237#discussion_r508943888) about using `network.GetServiceHostname`.

### DIFF
--- a/config/203-local-gateway.yaml
+++ b/config/203-local-gateway.yaml
@@ -57,7 +57,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: knative-local-gateway
-  namespace: knative-serving
+  namespace: istio-system
   labels:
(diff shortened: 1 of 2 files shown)


tcnghia:
@ZhiminXiang can you please help review this?  Thanks

PR #394, opened 2020-10-30
Run net-istio tests on KinD with non-default cluster suffix 

nak3:
This patch adds net-istio tests for KinD on github action.
The test also uses random cluster suffix.

Fixes https://github.com/knative-sandbox/net-istio/issues/343 https://github.com/knative-sandbox/net-istio/issues/344 

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -136,9 +136,7 @@ jobs:
         # Build and Publish our containers to the docker daemon (including test assets)
         export GO111MODULE=on
         export GOFLAGS=-mod=vendor
-        ko resolve -f test/config/ -f config/ | \
-          sed 's/imagePullPolicy:/# DISABLED: imagePullPolicy:/g' | \
-          kubectl apply -f -
+        ko resolve -f test/config/ -f config/ | kubectl apply -f -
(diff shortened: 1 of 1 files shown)


tcnghia:
Very nice! Thanks @nak3 !

.github/workflows/kind-e2e.yaml
```diff
@@ -0,0 +1,185 @@
...
+    - name: Create KinD Cluster
+      run: |
+        set -x
+
+        # KinD configuration.
+        cat > kind.yaml <<EOF
+        apiVersion: kind.x-k8s.io/v1alpha4
+        kind: Cluster
+        nodes:
+        - role: control-plane
```
JRBANCEL, line 100: Are two nodes enough to run the E2E tests? In `serving`, I had to add more nodes, see https://github.com/knative/serving/blob/master/.github/workflows/kind-e2e.yaml#L139

nak3: Yes, as far as I tested, the conformance test for `networking(net-istio)` does not use much resources like `serving`. We can add extra nodes if the tests start failing in the future.

tcnghia (approved):
/lgtm

PR #402, opened 2020-11-04
Include commit ID in component YAML instead of release YAML.

JRBANCEL:
In https://github.com/knative-sandbox/net-istio/pull/288, the commit was only added to `release.yaml`.

The problem is that in `serving` we copy the component YAML(s) directly, not `release.yaml`.

This includes the commit ID in all component YAML (which also end up in `release.yaml` by concatenation.

/assign @tcnghia


tcnghia (approved):
/lgtm

Issue #173, opened 2020-07-09
Run e2e test mesh mode with "mesh only" config and without cluster-local-gateway


nak3:
As https://github.com/knative/serving/issues/8585 reported, "mesh only" setting has some simple bug and e2e tests missed them.

We should run e2e test mesh mode with `local-gateway.mesh: "mesh"` and without cluster-local-gateway.

PR #405, opened 2020-11-09
Use mesh mode when sidecar injection is enabled

nak3:
This is a part of https://github.com/knative-sandbox/net-istio/issues/173.

By this patch, it enabled [`MeshOnly` API](https://github.com/knative/networking/blob/59e3a403ce6e06847bd610126cc2a66565fb9fd7/pkg/apis/networking/v1alpha1/ingress_types.go#L337-L339)
in ingress status.

### DIFF
--- a/third_party/istio-latest/install-istio.sh
+++ b/third_party/istio-latest/install-istio.sh
@@ -33,7 +33,7 @@ tar xzf ${ISTIO_TARBALL}
 # Enable mTLS STRICT in mesh mode
 if [[ $MESH -eq 1 ]]; then
   kubectl apply -f "$(dirname $0)/extra/global-mtls.yaml"
-  kubectl patch configmap/config-istio -n knative-serving --patch='{"data":{"local-gateway.mesh":"mesh"}}'
+  kubectl patch configmap/config-istio -n ${SYSTEM_NAMESPACE} --patch='{"data":{"local-gateway.mesh":"mesh"}}'
 fi
 
 # Clean up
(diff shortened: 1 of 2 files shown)


nak3:
https://github.com/knative/serving/pull/10064 verified it does not break it.

third_party/istio-latest/install-istio.sh
```diff
@@ -18,6 +18,7 @@
 ISTIO_VERSION=1.7.3
 ISTIO_TARBALL=istio-${ISTIO_VERSION}-linux-amd64.tar.gz
 DOWNLOAD_URL=https://github.com/istio/istio/releases/download/${ISTIO_VERSION}/${ISTIO_TARBALL}
+SYSTEM_NAMESPACE="${SYSTEM_NAMESPACE:-"knative-serving"}"
```
ZhiminXiang, line 21: why do we need to have a different system namespace? is it because we want to have a new namespace without cluster local gateway installed?

nak3: Thank you. (I made a mistake in `kubectl patch` below and fixed.)

This is because serving's e2e test deploys `config-istio` in the `SYSTEM_NAMESPACE` namespace 
https://github.com/knative/serving/blob/cbcd2ba4839711eafe1f846b6d886c84e8f56058/test/e2e-networking-library.sh#L62-L63 so we need to modify the `config-istio` in the namespace.

tcnghia (approved):
/lgtm

PR #407, opened 2020-11-10
Replace sample-controller with net-istio in CONTRIBUTING.md

nak3:
This patch makes a tiny change to replace `sample-controller` with
`net-istio` in CONTRIBUTING.md.

/cc @ZhiminXiang @JRBANCEL @arturenault


markusthoemmes (approved):
/lgtm
/approve

We might want to change this in sample-controller to not be copied around.

PR #412, opened 2020-11-12
Rename file to comply with divisive language

markusthoemmes:
Dummy is considered divisive language and the alternative is much better anyway.

/assign @nak3


nak3 (changes requested):
Could you rename this file https://github.com/knative-sandbox/net-istio/blob/master/third_party/istio-stable/dummy.go as weell?

It is necessary for docs repo.

markusthoemmes:
Will do, third_party is ignored by the checker by default.

/hold

nak3 (approved):

PR #445, opened 2020-12-08
Update istio-latest to 1.8.1

arturenault:
# Changes

- Update istio-latest to 1.8.1.

### DIFF
--- a/third_party/istio-latest/istio-ci-mesh.yaml
+++ b/third_party/istio-latest/istio-ci-mesh.yaml
@@ -43,8 +43,6 @@ spec:
   addonComponents:
     pilot:
       enabled: true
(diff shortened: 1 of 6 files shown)


third_party/istio-latest/istio-ci-mesh.yaml
```diff
@@ -43,8 +43,6 @@ spec:
   addonComponents:
     pilot:
       enabled: true
-    prometheus:
```
ZhiminXiang, line 46: out of curiosity, why does this need to be deleted?

arturenault: details here: https://github.com/knative/serving/issues/10221

PR #469, opened 2020-12-24
Fix tiny logging/comment/naming bugs in virtualservice status test

arturenault:
- Rename function from `virtualServiceCalledTimes` to `proberCalledTimes`, which is more accurate
- Fix a comment and an error log that both assumed n always == 1.


vagababov (commented):
/lgtm
/approve

PR #471, opened 2020-12-29
Move istio config patches into a separate file

arturenault:
# Changes

- Move bad config patches into their own file. They don't currently work because they're applied before  the configmap is created.


arturenault:
Moved changes into a script, so this one should be merged first now.

/assign @JRBANCEL @nak3 
/unhold

PR #498, opened 2021-01-28
Fix mesh tests

arturenault:
- knative/test-infra#2641 enabled mesh tests on pull requests in net-istio, but net-istio didn't support that parameter yet. This should fix that.


nak3:
/lgtm cancel

@arturenault probably net-istio needs to add injection label to the test namespace like this - https://github.com/knative/serving/blob/master/test/e2e-common.sh#L420-L423 ?

arturenault:
@nak3 you're right. I need to learn to read bash properly; I didn't see there were two different functions called `test_setup`, I thought net-istio was just using serving's.

Sorry for the noise and extra work!

nak3:
/lgtm
/test pull-knative-sandbox-net-istio-latest-mesh

No problem :+1:

tcnghia (approved):

PR #499, opened 2021-01-28
peer auth: Only set permissive mTLS on the webhook port.

nairb774:
# Changes

Rather than making mTLS permissive on all ports, only open up the one
port that needs to be accessed by the control plane. This places the
metrics and profiling endpoints back under mTLS being required.

This technically enables the minimal carve out needed, and has been how
we have been running the KNative webhooks successfully.

/kind enhancement

From my perspective it technically improves security - at least as far as auditing is concerned.

The purpose is to reduce the surface area of things running in the cluster that is not covered by mTLS. Whitelisting one port is better than whitelisting the whole service.

**Release Note**

```release-note
Make mTLS permissive only on the webhook port. Other ports will go back to using the cluster default setting.
```

**Docs**

N/A?

### DIFF
--- a/config/400-webhook-peer-authentication.yaml
+++ b/config/400-webhook-peer-authentication.yaml
@@ -12,8 +12,9 @@ spec:
   selector:
     matchLabels:
       app: webhook
-  mtls:
-    mode: PERMISSIVE
+  portLevelMtls:
+    8443:
+      mode: PERMISSIVE
 ---
 apiVersion: "security.istio.io/v1beta1"
 kind: "PeerAuthentication"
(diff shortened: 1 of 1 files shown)


nairb774:
Sorry - default email and CLA email are different and I always forget for Google repos.

nairb774:
Test failure looks unrelated (?):

> 2021/01/29 02:22:04 Error during command execution: unknown flag: --mesh

nak3:
Yeah, https://github.com/knative-sandbox/net-istio/pull/498 is fixing the issue now.

tcnghia (approved):

PR #263, opened 2020-09-04
Update dependencies.

JRBANCEL:
There are a lot of changes upstream breaking the repo.

The only thing I can't figure out is why some AutoTLS tests are seeing the wrong Gateway being created.
@ZhiminXiang  do you have any clue?

PR #508, opened 2021-02-05
Re-introduce behaviour of explicitly setting retries to 0

julz:
# Changes

- :bug: Re-introduces the behaviour of explicitly setting retries to zero, needed to override the Istio default of 2 retries. This was introduced in https://github.com/knative-sandbox/net-istio/pull/159 to fix https://github.com/knative/serving/issues/6549, but looks like it was accidentally dropped in https://github.com/knative-sandbox/net-istio/pull/263 when we bumped deps after removing the Retry field from KIngress.

/kind bug

**Release Note**

```release-note
Configures Istio not to retry requests. This was accidentally reverted in 0.18.
```

/assign @tcnghia @nak3

### DIFF
--- a/pkg/reconciler/ingress/resources/virtual_service.go
+++ b/pkg/reconciler/ingress/resources/virtual_service.go
@@ -214,6 +214,7 @@ func makeVirtualServiceRoute(ctx context.Context, hosts sets.String, http *v1alp
 	}
 
 	route := &istiov1alpha3.HTTPRoute{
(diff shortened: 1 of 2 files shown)


duglin:
testing it but if this fixes our issue, can we get a test added to ensure this isn't removed by mistake again?

julz:
> testing it but if this fixes our issue, can we get a test added to ensure this isn't removed by mistake again?

Working on adding a test to [networking/conformance](https://github.com/knative/networking/tree/master/test/conformance) right now

nak3 (approved):
/lgtm

PR #512, opened 2021-02-09
Properly implement Splits support in RewriteHost

julz:
We added support for specifying (actually, a requirement to supply!) splits to the API (https://github.com/knative/networking/pull/107), but the istio implementation was still actually hard-coding the gateway as the backend for RewriteHost rules, and ignoring the splits. 

This changes that so we properly respect the passed splits like the other ingresses do. (In practice none of our current use cases work differently either way so this didn't matter, but that could change and it's good to be consistent).

/kind enhancement

**Release Note**

```release-note
Ingresses with RewriteHost specified now properly respect passed Splits
```

/assign @nak3 @tcnghia


nak3:
LGTM except for Lint issue https://github.com/knative-sandbox/net-istio/pull/512/checks?check_run_id=1863586626

nak3 (approved):
/lgtm

PR #516, opened 2021-02-11
Bump Golang version to 1.15

markusthoemmes:
As per title. Other repos have long moved.

/assign @nak3


vagababov:
It's time we moved to 1.16 :/

PR #524, opened 2021-02-19
Add placeholder.go back to fix docs

mattmoor:


tcnghia (approved):
/lgtm

csantanapr:
Any comments why is this the fix? I want to learn for next time

mattmoor:
@csantanapr You can only vendor directories with Go code using `go mod`, so without the Go file (which was dropped) docs cannot vendor these config files.

Issue #514, opened 2021-02-10
Fail to create TLSGateway from DomainMapping's Kingress due to invalid name


nak3:
### Description

`serving-https test` is always failing on `TestDomainMapping`. Please refer to any [log](https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-https/1359437522937057280) in `https` test in [test grid](https://testgrid.k8s.io/knative-serving#knative-serving-https).

This is because net-istio tries to create gateway name with "dot" like `domain-mapping-hfyjwokc.example.org-3797421420`.

```
serving-tests                          39m         Warning   InternalError                  ingress/domain-mapping-hfyjwokc.example.org                                             admission webhook "validation.istio.io" denied the request: configuration is invalid: invalid gateway name: "domain-mapping-hfyjwokc.example.org-3797421420"
```

#### For more detail...

- Gateway name is created from KIngress's name:

https://github.com/knative-sandbox/net-istio/blob/3bba898d0fed6640de929e3ff8f8bcaddb66bdee/pkg/reconciler/ingress/resources/gateway.go#L265-L268

- Istio does not allow dots.

https://github.com/istio/istio/blob/b9545357f0d57ed31d24a8ad8ea7a3cea883f63a/pkg/config/validation/validation.go#L334-L337

### Steps to produce

- Please refer to https://testgrid.k8s.io/knative-serving#knative-serving-https
- Or you can re-create the issue with autoTLS & DomainMapping.

ZhiminXiang:
So the Gateway name has dot because the name of the Kingress has dot. 
Maybe we should check if `accessor.GetName()` is valid. If it is invalid, we replace it with a valid format.

tcnghia:
Fixed by #532

PR #532, opened 2021-03-02
Handle the case when ingress name has dot

ZhiminXiang:
Fixes #
https://github.com/knative-sandbox/net-istio/issues/514

**Release Note**

```release-note
Fix the incorrect Gateway name format for DomainMapping auto TLS feature for Istio implmenetation

```

### DIFF
--- a/pkg/reconciler/ingress/resources/gateway.go
+++ b/pkg/reconciler/ingress/resources/gateway.go
@@ -203,7 +203,7 @@ func makeWildcardGateways(ctx context.Context, originWildcardSecrets map[string]
 // if this function is changed in the upstream (for example, Istio allows the dot in the future), we don't want to
 // import the change without awareness because it could break the compatibility of Gateway name generation.
 func isDNS1123Label(value string) bool {
-	return len(value) <= dNS1123LabelMaxLength && dns1123LabelRegexp.MatchString(value)
+	return len(value) <= dns1123LabelMaxLength && dns1123LabelRegexp.MatchString(value)
 }
 
 // WildcardGatewayName creates the name of wildcard Gateway.
@@ -323,7 +323,7 @@ func MakeTLSServers(ing *v1alpha1.Ingress, ingressTLS []v1alpha1.IngressTLS, gat
 	return SortServers(servers), nil
 }
 
-func portNamePrefix(prefix, suffx string) string {
+func portNamePrefix(prefix, suffix string) string {
 	if !isDNS1123Label(suffix) {
 		suffix = fmt.Sprint(adler32.Checksum([]byte(suffix)))
 	}
--- a/pkg/reconciler/ingress/resources/gateway.go
+++ b/pkg/reconciler/ingress/resources/gateway.go
@@ -44,7 +44,7 @@ import (
 // GatewayHTTPPort is the HTTP port the gateways listen on.
 const (
 	GatewayHTTPPort       = 80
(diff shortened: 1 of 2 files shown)


pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -189,6 +196,16 @@ func makeWildcardGateways(ctx context.Context, originWildcardSecrets map[string]
 	return gateways, nil
 }
 
+// IsDNS1123Label tests for a string that conforms to the definition of a label in
+// DNS (RFC 1123).
+// This function is copied from https://github.com/istio/istio/blob/806fb24bc121bf93ea06f6a38b7ccb3d78d1f326/pkg/config/labels/instance.go#L97
+// We directly copy this function instead of importing it into vendor and using it because
+// if this function is changed in the upstream (for example, Istio allows the dot in the future), we don't want to
+// import the change without awareness because it could break the compatibility of Gateway name generation.
+func isDNS1123Label(value string) bool {
```
JRBANCEL, line 205: Can we use the existing https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L187 ?

ZhiminXiang: I think technically we can either use istio one or k8s one.

But I am inclined to copy the function into this repo because if there is any change to this function in the upstream (e.g. allowing dot in the future), then the gateway name could be changed implicitly and we may not be aware of this.
If the version of the function we use in the our code is not compatible with the requirement of Istio or k8s in the future, then our E2E tests should be failed and we will get notice about it. WDYT?

JRBANCEL: I get your point, but this is not some random function, it follows RFC 1123. It is not going to change.
Up to you.

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -304,6 +323,14 @@ func MakeTLSServers(ing *v1alpha1.Ingress, ingressTLS []v1alpha1.IngressTLS, gat
 	return SortServers(servers), nil
 }
 
+func portNamePrefix(prefix, ingressName string) string {
```
nak3, line 326:
```suggestion
func portNamePrefix(ingressNamespace, ingressName string) string {
```

or 

```suggestion
func portNamePrefix(prefix, suffix string) string {
```

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -41,7 +42,11 @@ import (
 )
 
 // GatewayHTTPPort is the HTTP port the gateways listen on.
-const GatewayHTTPPort = 80
+const (
+	GatewayHTTPPort       = 80
+	DNS1123LabelMaxLength = 63 // Public for testing only.
```
nak3, line 47:
```suggestion
	dns1123LabelMaxLength = 63
```

I think this is not used for testing :thinking:

nak3 (commented):
Left a few comments but looks good to me overall.

pkg/reconciler/ingress/resources/gateway.go
ZhiminXiang: Done

pkg/reconciler/ingress/resources/gateway.go
ZhiminXiang: Done

nak3 (approved):

PR #541, opened 2021-03-10
Use prefix matching for KIngress rule paths

impl:
:broom: This change updates the matching mode for rule paths to use prefix matching instead of regular expression matching. The regular expression mode was never correctly implemented by Knative when generating rules, and most other networking layers only support prefix matching anyway.

See the discussion here for more context: https://github.com/knative/networking/pull/364

/kind cleanup


tcnghia:
/ok-to-test
cc @JRBANCEL @arturenault @ZhiminXiang

arturenault (approved):
/lgtm

ZhiminXiang (approved):

PR #546, opened 2021-03-11
Add the reconciler machinery for pod addressability

markusthoemmes:
# Changes

Ref knative/serving#10751

This is just creating the initial scaffolding with the expected resources so that subsequent diffs are more relevant to review.

/assign @julz @arturenault


pkg/reconciler/serverlessservice/controller.go
```diff
@@ -0,0 +1,83 @@
...
+
+	c := &reconciler{
+		kubeclient:            kubeclient.Get(ctx),
+		istioclient:           istioclient.Get(ctx),
+		virtualServiceLister:  virtualServiceInformer.Lister(),
+		destinationRuleLister: destinationRuleInformer.Lister(),
+	}
+	impl := sksreconciler.NewImpl(ctx, c, func(impl *controller.Impl) controller.Options {
+		logger.Info("Setting up ConfigMap receivers")
+		resyncIngressesOnConfigChange := configmap.TypeFilter(&config.Istio{})(func(string, interface{}) {
```
julz, line 56: is this variable name correct here? I think it should be sthg like `resyncSKSOnConfigChange`

markusthoemmes: Too much copy and paste... thanks!

julz (approved):
/lgtm

vagababov (approved):
/lgtm

PR #558, opened 2021-03-16
Add an accessor for DestinationRule objects

markusthoemmes:
# Changes

Ref knative/serving#10751

Adds a helper to reconcile destionationrules the same way we have an accessor for virtualservices. The code is mostly copy and pasted and then the respective resource names have been replaced.

/assign @vagababov @ZhiminXiang

### DIFF
--- a/pkg/reconciler/accessor/istio/destinationrule.go
+++ b/pkg/reconciler/accessor/istio/destinationrule.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Knative Authors
+Copyright 2021 The Knative Authors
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/accessor/istio/destinationrule.go
```diff
@@ -0,0 +1,86 @@
+/*
+Copyright 2019 The Knative Authors
```
ZhiminXiang, line 2: should this be 2021?

markusthoemmes: Indeed, fixed!

JRBANCEL (approved):

PR #560, opened 2021-03-17
fix name conflict in image registry and add webhook port env

MartinForReal:
# Changes

- bug: Fix bug https://github.com/knative/operator/issues/519

/kind bug

Fixes https://github.com/knative/operator/issues/519

**Release Note**

```release-note
fix installation issue using operator when images are pulled from private registry
```

### DIFF
--- a/config/webhook-deployment.yaml
+++ b/config/webhook-deployment.yaml
@@ -36,7 +36,7 @@ spec:
     spec:
       serviceAccountName: controller
       containers:
-      - name: webhook
+      - name: istio-webhook
         # This is the Go import path for the binary that is containerized
         # and substituted here.
         image: ko://knative.dev/net-istio/cmd/webhook
@@ -64,7 +64,8 @@ spec:
           value: knative.dev/net-istio
         - name: WEBHOOK_NAME
           value: istio-webhook
-
+        - name: WEBHOOK_PORT
+          value: '8443'
         securityContext:
           allowPrivilegeEscalation: false
 


config/webhook-deployment.yaml
```diff
@@ -36,7 +36,7 @@ spec:
     spec:
       serviceAccountName: controller
       containers:
-      - name: webhook
+      - name: istio-webhook
```
nak3, line 39: Is this necessary to change?

As per operator's docs https://github.com/knative/operator/blob/main/docs/configuration.md#specregistryoverride, 
> If the container name is not unique across all of the deployments, daemonsets and jobs, you can prefix the container name with the "parent's" name and a slash, e.g. deployment/container.

you can avoid the conflict by `webhook/webhook`  and `net-istio/webhook`?

nak3: If we changed the container name, current users' clusters who uses `net-istio/webhook` will be broken after the upgrade.

MartinForReal: if we changed the name of webhook container in webhook deployment, we will trigger an update for deployment resource. I don't think this change will ends up with an service outage.

MartinForReal: > you can avoid the conflict by `webhook/webhook` and `net-istio/webhook`?

As is quoted below, container name should be a DNS_LABEL
> The name of this port within the service. This must be a DNS_LABEL. All ports within a ServiceSpec must have unique names. When considering the endpoints for a Service, this must match the 'name' field in the EndpointPort. Optional if only one ServicePort is defined on this service.
[ref](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#container-v1-core)

MartinForReal: How about change the name of the container to net-istio-webhook?

nak3: Please let me make it clear.
You are trying to fix https://github.com/knative/operator/issues/519, correct?

The root cause of https://github.com/knative/operator/issues/519 is `spec.registry.overrride.webhook` below:

```
spec:
  registry:
    override:
      webhook: >-
      knative/webhook:fdf57643768d1b4337dbb347dade96cbdb2cdca74e564909f73895bf377fbadf
```

where the container name `webhook` conflicts with serving's webhook and net-istio's webhook.
Then, operator prepares the solution for the problem by configure `deployment/container` like:

```
spec:
  registry:
    override:
      webhook/webhook: >-
      knative/webhook:fdf57643768d1b4337dbb347dade96cbdb2cdca74e564909f73895bf377fbadf
      net-istio/webhook: >-
      knative/net-istio:somethingsomething-hash
```
I think it is just an operator's configuration so [ref](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#container-v1-core) should not be related.

Finally, if some existing users already used the solution below, then changing the container name breaks it.

```
spec:
  registry:
    override:
      net-istio/webhook: knative/net-istio:someimportantimage-hash
```

MartinForReal: I see, you are right. Thanks for clarification!

MartinForReal:
After I change the name to `istio-webhook/webhook`, webhook is created successfully. Thanks for the help!

PR #564, opened 2021-03-22
Start pod addressability reconciler with the controller

markusthoemmes:
# Changes

Ref https://github.com/knative/serving/issues/10751

This enables the new reconciler and adjusts the necessary ClusterRole. Note that the reconciler is disabled by default, i.e. it'll do nothing.

/assign @nak3 @julz @ZhiminXiang @vagababov

### DIFF
--- a/test/config/300-serverlessservice.yaml
+++ b/test/config/300-serverlessservice.yaml
@@ -1,0 +1,66 @@
+# Copyright 2019 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


julz (dismissed):
lgtm

nak3 (approved):
/lgtm

mesh test failed but it seems unrelated issue.

test/config/300-serverlessservice.yaml
```diff
@@ -0,0 +1,66 @@
+# Copyright 2019 The Knative Authors
```
vagababov, line 1: 2021.
Is there anything we need to keep in sync?

markusthoemmes: Already sent the respective PRs to clean up the copied CRDs.

PR #553, opened 2021-03-15
Add configuration for graceful shutdown to ingressgateways

markusthoemmes:
# Changes

We seem to be seeing flaky connections in the Serving E2E test suite due to ingressgateways shutting down during tests. We're configuring this value for mesh pods as well, so maybe this'll help (the default value is 5s).

/hold

For verification in Serving first.

PR #570, opened 2021-03-25
Disable autoscaling for ingress gateways

julz:
HPA scaling down ingress gateways causes flakes in CI, which reflect legitimate problems production users might experience.

We tried [bumping the drain time to 20 seconds](https://github.com/knative-sandbox/net-istio/pull/553) but we still see flakes since one test requires at least 35 seconds. In reality this feels like a real problem unless we set the shutdown time to 5 minutes+ (the default max revision timeout), and even then scaling down an ingress gateway will kill long running requests (e.g. websockets).

Bumping the drain time again, to 40 seconds, would also fix the flakes, but would be covering up a real problem that would hit users in production. Disabling autoscaling for the ingress gateway seems like the safer out-of-the-box default for now.

# Changes

- :bug: Disable autoscaling of ingress gateways, since this interacts badly with long running requests

/kind bug

**Release Note**

```release-note
Disable autoscaling of ingress gateways out of the box, since this interacts badly with long running requests. Operators that wish to enable autoscaling of ingress gateways should make sure to set the drain timeouts high enough for their longest running requests.
```

/assign @markusthoemmes @ZhiminXiang @nak3

### DIFF
--- a/third_party/istio-head/istio-ci-mesh.yaml
+++ b/third_party/istio-head/istio-ci-mesh.yaml
@@ -30,8 +30,8 @@ spec:
         targetAverageUtilization: 60
     gateways:
       istio-ingressgateway:
-        autoscaleMin: 2
-        autoscaleMax: 5
+        autoscaleMin: 3
+        autoscaleMax: 3
     sidecarInjectorWebhook:
       rewriteAppHTTPProbe: true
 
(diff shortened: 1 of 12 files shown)


markusthoemmes (approved):
/lgtm
/approve
/hold

To leave time for networking folks to have a look as well.

ZhiminXiang:
@julz do you mind creating an issue in net-istio repo to track the ingress problem with what you found for the further debugging?

I think this PR is kind of mitigation. So better to have some track to help root causing it :)

ZhiminXiang (approved):

JRBANCEL (approved):

vagababov:
/test all

let's see if 3 is enough for meshes and stuff

markusthoemmes:
/retest

Surely unrelated

PR #594, opened 2021-04-13
Update istio manifest to make it simple

nak3:
Part of https://github.com/knative/docs/issues/3442

This patch removes the following configurations from manifest:

- `useMCP: false` ... not necessary to set it as the default is false.
- `jwtPolicy: first-party-jwt` ... As per it does not work in k8s 1.21 - https://github.com/knative/docs/issues/3442
- `autoInject: disabled` ... not necessary to set - https://github.com/knative/docs/issues/3442
- `rewriteAppHTTPProbe: true` ... not necessary to set it as the default is true.
- `pilot.enabled: true` ... not necessary to set it as the default is true.
- `ingressGateways.enabled: true` ... not necessary to set it as the default is true.

/kind cleanup

**Release Note**

```release-note
istio.yaml drops some configurations. Especially removing `jwtPolicy: first-party-jwt` and `autoInject: disabled` from istio.yaml might impact on the cluster.
```

/cc @arturenault @dprotaso @julz


dprotaso (approved):

nak3:
/cc @tcnghia @markusthoemmes @ZhiminXiang 

Could you please approve this PR?

markusthoemmes (approved):
/lgtm
/approve

PR #595, opened 2021-04-13
install resources in a particular order

dprotaso:
This PR orders the resources applied to help mitigate the installation flakes when using `kubectl apply -f istio.yaml`

/assign @nak3 @arturenault

### DIFF
--- a/third_party/library.sh
+++ b/third_party/library.sh
@@ -123,11 +123,36 @@ function generate() {
   generate_manifests "$path" "$@"
 }
 
-function install_yaml() {
-  if grep -q "CustomResourceDefinition" "$1"; then
-    kubectl apply --selector="knative.dev/crd-install=true" -f "$1"
-    kubectl wait --for=condition=Established --all crd
-  fi
+function run_yq() {
+  run_go_tool github.com/mikefarah/yq/v4 yq "$@"
+}
 
+function install_yaml() {
+  local order=(
+    Namespace
+    ServiceAccount
+    ClusterRole
+    ClusterRoleBinding
+    Role
+    RoleBinding
+    CustomResourceDefinition
+    ConfigMap
+    Deployment
+    Service
+    HorizontalPodAutoscaler
+    PodDisruptionBudget
+    MutatingWebhookConfiguration
+    ValidatingWebhookConfiguration
+
+    # Istio resources
+    EnvoyFilter
+    PeerAuthentication
+  )
+
+  for kind in "${order[@]}"; do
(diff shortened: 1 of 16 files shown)


third_party/library.sh
```diff
@@ -123,11 +123,36 @@ function generate() {
...
 }
 
+function run_yq() {
+  run_go_tool github.com/mikefarah/yq/v4 yq "$@"
+}
+
 function install_yaml() {
-  if grep -q "CustomResourceDefinition" "$1"; then
-    kubectl apply --selector="knative.dev/crd-install=true" -f "$1"
-    kubectl wait --for=condition=Established --all crd
```
nak3, line 129: We will not wait for `Established` condition for CRD anymore. Is it alright?

dprotaso:
hmmm.. I think `run_go_tool` isn't working when the version is in the module path

third_party/library.sh
dprotaso: Yeah - i dropped it - I want to see if it flaked in the presubmit. N `kubectl` invocation is really slow so I figured it wouldn't come up

third_party/library.sh
```diff
@@ -123,11 +123,36 @@ function generate() {
...
+function run_yq() {
+  GO111MODULE=on run_go_tool github.com/mikefarah/yq/v4 yq "$@"
+}
+
 function install_yaml() {
-  if grep -q "CustomResourceDefinition" "$1"; then
-    kubectl apply --selector="knative.dev/crd-install=true" -f "$1"
-    kubectl wait --for=condition=Established --all crd
-  fi
+  local order=(
```
arturenault, line 131: this is cool. feels like there should be a way in kubectl to specify these sorts of orderings... but i guess that wouldn't be declarative

dprotaso: I think kubectl just passes things along as they appear in `yaml` - in serving we structure our config directory so the resulting yaml is ordered in such a way so it works most of the time.

The `istioctl` output is somewhat ordered but

nak3:
`kubectl apply` creates the resources in the order they appear in the file. So do we want to change the order in `istio.yaml` by this hack?
Then, users also can just run `kubectl apply -f istio.yaml`.

dprotaso:
> kubectl apply creates the resources in the order they appear in the file. So do we want to change the order in istio.yaml by this hack?
Then, users also can just run kubectl apply -f istio.yaml.

Done - we'll do both. I kept the logic in the install script since you could have races for resources in the same file.

dprotaso:
rebased

nak3 (approved):

PR #599, opened 2021-04-16
Use SKS' PrivateServiceName field rather than generate it

markusthoemmes:
# Changes

As per title, this makes the SKS reconciler here a little more robust and relies on the correct contracts.

/assign @nak3


nak3 (approved):
/lgtm

PR #600, opened 2021-04-16
Drop istio-minimal.yaml

nak3:
`istio-minimal.yaml (istio-local.yaml)` was introduced by https://github.com/knative/serving/pull/6025
for local deployments.
But today `istio-minimal.yaml` is equivalent to `istioctl install -y` so we
don't need to keep the manifest in the repo.

Note, current `istio-minimal.yaml` has
`TERMINATION_DRAIN_DURATION_SECONDS` but it was added to solve flake
of e2e test by https://github.com/knative-sandbox/net-istio/pull/553 so it is not mandatory.

/kind cleanup

/cc @markusthoemmes @arturenault @dprotaso


arturenault (dismissed):
nice, much cleaner.

/lgtm

arturenault (approved):
/lgtm

markusthoemmes (approved):
/lgtm
/approve

PR #517, opened 2021-02-11
Temporarily disable mTLS STRICT in mesh mode.

JRBANCEL:
See #503.
Let's see if that is the issue.

/assign @nak3 @arturenault

Issue #503, opened 2021-02-02
mTLS with STRICT mode is not stable


nak3:
Since enabling mTLS strict mode by fixing a bug by https://github.com/knative/serving/pull/10635,
some tests became unstable. 

Please refer to https://testgrid.knative.dev/serving#istio-stable-mesh. `TestScaleToN` test had been unstable for long time, but other tests became red after the change https://github.com/knative/serving/pull/10635.

PR #617, opened 2021-04-30
[istio-stable] renable global mTLS when in mesh mode

dprotaso:
mTLS was disabled in https://github.com/knative-sandbox/net-istio/pull/517  But this is causing confusion downstream in `serving` since we didn't realize the change had occurred.

Going forward I think we should have consistency with our `istio-*` yamls and perform any specific overrides in `serving` 

Related: https://github.com/knative-sandbox/net-istio/issues/503

### DIFF
--- a/third_party/istio-stable/extra/istio-mesh.yaml
+++ b/third_party/istio-stable/extra/istio-mesh.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: "security.istio.io/v1beta1"
+kind: "PeerAuthentication"
+metadata:
+  name: "default"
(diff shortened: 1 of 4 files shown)


nak3:
/lgtm
/assign @arturenault 

> Going forward I think we should have consistency with our istio-* yamls and perform any specific overrides in serving

Yes, agree.

arturenault (approved):
/lgtm

Great idea. When we first put in mTLS strict we were still generating templates on the fly, this is much cleaner.

nak3 (approved):

PR #620, opened 2021-05-11
Add PILOT_ENABLED_SERVICE_APIS to manifest

nak3:
Istio needs this option to support Gateway API (Ingress v2).
If istio manifest could enable the option by default, serving repo
will add Gateway API support with the manifest in net-istio wihtout
change.

/cc @arturenault @dprotaso @markusthoemmes

### DIFF
--- a/third_party/istio-head/istio-ci-mesh.yaml
+++ b/third_party/istio-head/istio-ci-mesh.yaml
@@ -24,6 +24,8 @@ spec:
       autoscaleMax: 10
       cpu:
         targetAverageUtilization: 60
+      env:
+        PILOT_ENABLED_SERVICE_APIS: true
     gateways:
       istio-ingressgateway:
         autoscaleEnabled: false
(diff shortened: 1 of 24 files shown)


arturenault (approved):
/lgtm

third_party/istio-head/istio-ci-mesh.yaml
```diff
@@ -24,6 +24,8 @@ spec:
       autoscaleMax: 10
       cpu:
         targetAverageUtilization: 60
+      env:
+        PILOT_ENABLED_SERVICE_APIS: true
```
howardjohn, line 28: this is default in latest istio fwiw

nak3: Oh, thank you! Istio v1.10+ enabled it by default.Then, we should update to 1.10 by https://github.com/knative-sandbox/net-istio/issues/631 first then add this option in 1.9 only.

nak3:
Closing as 1.10 enabled the option by default and istio-latest uses 1.10.1 now.

Issue #630, opened 2021-05-18
Optimize number of Envoy routes created for a single KService


howardjohn:
When creating a single kservice, we end up with a lot of Envoy routes. For example, on the local gateway:
```
http.8081     hello.default.172.17.255.155.sslip.io     /*                     hello-ingress.default
http.8081     hello.default.172.17.255.155.sslip.io     /*                     hello-ingress.default
http.8081     hello.default.172.17.255.155.sslip.io     /*                     hello-ingress.default
http.8081     hello.default.172.17.255.155.sslip.io     /*                     hello-ingress.default
http.8081     hello.default.172.17.255.155.sslip.io     /*                     hello-ingress.default
http.8081     hello.default.172.17.255.155.sslip.io     /*                     hello-ingress.default
http.8081     hello.default.svc.cluster.local           /*                     hello-ingress.default
http.8081     hello.default.svc.cluster.local           /*                     hello-ingress.default
http.8081     hello.default.svc.cluster.local           /*                     hello-ingress.default
http.8081     hello.default.svc.cluster.local           /*                     hello-ingress.default
http.8081     hello.default.svc.cluster.local           /*                     hello-ingress.default
http.8081     hello.default.svc.cluster.local           /*                     hello-ingress.default
http.8081     hello.default.svc                         /*                     hello-ingress.default
http.8081     hello.default.svc                         /*                     hello-ingress.default
http.8081     hello.default.svc                         /*                     hello-ingress.default
http.8081     hello.default.svc                         /*                     hello-ingress.default
http.8081     hello.default.svc                         /*                     hello-ingress.default
http.8081     hello.default.svc                         /*                     hello-ingress.default
http.8081     hello.default                             /*                     hello-ingress.default
http.8081     hello.default                             /*                     hello-ingress.default
http.8081     hello.default                             /*                     hello-ingress.default
http.8081     hello.default                             /*                     hello-ingress.default
http.8081     hello.default                             /*                     hello-ingress.default
http.8081     hello.default                             /*                     hello-ingress.default
```

We get 24 routes. This is 4 variants of the domain X override/no override X 3 variants of the domain (excluding the external one).

I have been looking mostly into the internal gateway, but the external one has similar issues, although a few less matches it seems.

I am investigating this more, opening this issue so I can track what I learn

howardjohn:
One major issue is the matches:
```
  - match:
    - authority:
        prefix: hello.default
      gateways:
      - knative-serving/knative-local-gateway
[...]

PR #632, opened 2021-05-19
Deduplicate host matches to avoid excessive matches

howardjohn:
For  https://github.com/knative-sandbox/net-istio/issues/630

# Changes

This reduces the number of routes generated for a single KService by 1/3
in most cases. I profiled this on the Istio control plane side and saw a
corresponding ~50% drop in CPU usage to compute the routes.

/kind performance

**Release Note**

```release-note
Optimized generated routes to minimize Envoy configuration size
```

**Docs**

```docs

```

### DIFF
--- a/pkg/reconciler/ingress/resources/virtual_service.go
+++ b/pkg/reconciler/ingress/resources/virtual_service.go
@@ -19,7 +19,6 @@ package resources
 import (
 	"context"
 	"fmt"
-	"sort"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -153,7 +154,11 @@ func makeVirtualServiceSpec(ing *v1alpha1.Ingress, gateways map[v1alpha1.Ingress
...
 	matches := []*istiov1alpha3.HTTPMatchRequest{}
 	clusterDomainName := network.GetClusterDomainName()
-	for _, host := range hosts.List() {
+
+	// Deduplicate hosts to avoid excessive matches, which cause a combinatorial expansion in Istio
+	distinctHosts := getDistinctHostPrefixes(hosts)
+
+	for _, host := range distinctHosts.List() {
 		g := gateways[visibility]
 		if strings.HasSuffix(host, clusterDomainName) && len(gateways[v1alpha1.IngressVisibilityClusterLocal]) > 0 {
```
howardjohn, line 163: One thing I am not sure of, which is causing tests to fail: This suffix check is failing now, since we are deduplicating the full hostname out. However, I don't see how it ever worked before.

For example, if I have a KIngress with `hosts: [a.default, a.default.svc, a.default.svc.cluster.local]; visibility: ExternalIP`, then we would decide that only the last host should be ClusterLocal, when in reality they are all aliases for the same host so they should probably all have the same visibility.

I would consider this a bug, and remove this detection and assume that rules have the proper visibility set. However, I am not sure the implications of this, so I left it as is. This causes the tests to fail so some advice is needed.

pkg/reconciler/ingress/resources/virtual_service.go
```diff
@@ -212,6 +217,32 @@ func makeVirtualServiceRoute(hosts sets.String, http *v1alpha1.HTTPIngressPath,
 	return route
 }
 
+// getDistinctHostPrefixes deduplicate a set of prefix matches. For example, the set {a, aabb} can be
+// reduced to {a}, as a prefix match on {a} accepts all the same inputs as {a, aabb}.
+func getDistinctHostPrefixes(hosts sets.String) sets.String {
+	all := hosts.List()
+	// First we sort the list. This ensures that we always process the smallest elements (which match against
+	// the most patterns, as they are less specific) first.
+	sort.Strings(all)
```
markusthoemmes, line 226: `hosts.List()` is already sorted. `UnsortedList` would be the unsorted version.

markusthoemmes:
This looks super promising and I wonder what this'll do to our scale testing! Thanks @howardjohn 

/assign @nak3 

Can you help with the question Kenjiro?

pkg/reconciler/ingress/resources/virtual_service.go
howardjohn: Thanks! I updated the tests to accommodate for this new behavior

nak3 (approved):
/lgtm

Thank you!

PR #636, opened 2021-05-24
Set "disable-gateway-port-translation" label on local gateway Service

howardjohn:
This was added to Istio in https://github.com/istio/istio/pull/33021. It
will likely be available in 1.9.next and 1.10.next, but if the label is
present with an older Istio version its not a problem - it will just not
have any impact.

What this does is avoid the ambiguities we currently have with the
Gateway service selection. Currently, if the local gateway is created
before the external gateway, we will be broken because the "port 80"
Gateway will match the local gateway and get translated to port 8081,
when it should be translated to port 8080 of the external gateway.

What this label does is tell Istio that the local gateway service is not
eligible for this port translation; as a result, Gateways on port 80
will never be mapped to port 8081. Instead, we expect that port 8081
will be used explicitly in the Gateway, as we do here.

tl;dr:
* Before: NACK if services created in wrong order
* After: service order does not matter

Note: the "experimental" label may be a red flag - this was intended to
be a scary label to dissuade any *other* than Knative from using it. The
label was made specifically to support Knative.

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/config/203-local-gateway.yaml
+++ b/config/203-local-gateway.yaml
@@ -42,6 +42,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
     networking.knative.dev/ingress-provider: istio
+    experimental.istio.io/disable-gateway-port-translation: "true"
 spec:
   type: ClusterIP
   selector:


nak3:
/cc @arturenault @julz @dprotaso 
Please let us know if there are any objections to add this label.

arturenault (approved):
/lgtm

julz (commented):
/lgtm

Thanks @howardjohn!

nak3 (approved):

dprotaso:
@howardjohn when does the feature flag switch and we'll be able to drop this label?

howardjohn:
Its not so much a feature flag - the label is basically enabling a non-standard behavior specifically to support knative doing a non-standard thing because we don't want to break knative.

Once http://gateway-api.org/ is used this problem will go away entirely (see https://docs.google.com/document/d/1gDhj058q2PKKI9SWpxE23ExEXdOkLiueT8JA_oEJ8Mo/edit# for long context on why); the changes we are making there *may* be backported to the standard Istio Gateway API but there are no current plans to do so. If we did it would be 1.11+. Do you typically allow multiple versions of Istio with a single Knative version?

dprotaso:
> Do you typically allow multiple versions of Istio with a single Knative version?

There's tests against different versions - ie. https://github.com/knative-sandbox/net-istio/pull/648

But we've been shipping what we deem 'stable' in our release yamls

ZhiminXiang:
Does this mean when installing Istio to support Knative, we need a specific installation instruction? Here is out current installation instruction https://github.com/knative/docs/blob/mkdocs/docs/admin/install/installing-istio.md which seems does not include this  extra label.

howardjohn:
No, the label is on the knative-local-gateway service which is installed by knative

Issue #637, opened 2021-05-25
Support individual Ingress's HTTPOption


nak3:
https://github.com/knative/networking/issues/301 tries to allow a specific Ingress to change http option but current net-istio uses one http server globally with `*`

https://github.com/knative-sandbox/net-istio/blob/ab9dfc86f07f3ce0da06f3cc42282fcf2e0ad82c/config/202-gateway.yaml#L27-L33

So one Ingress's HTTPOption effects on all HTTP behavior. In other words, net-istio have to use same HTTPOption within a cluster.

ZhiminXiang:
For the short term to support the backward compatibility, we can do the following works:
1. the global Gateway `knative-ingress-gateway` with HTTP protocol could be considered as the default Gateway when Kingress HTTPOption is HTTP. That is, all HTTP Kingress will be tied to the global Gateway.
2. When Kingress HTTPOption is not HTTP (e.g. HTTPS), we create a Kingress specific Gateway, and tie the Kingress to the Gateway.
This is what we are currently doing when auto-TLS is enabled.

For the long term, we could consider getting rid of the global Gateway (but TBH this could be difficult because in our current instructions we use the global Gateway as the API to manually configure TLS at this point).

PR #638, opened 2021-05-25
Use spec.HTTPOption in KIngress instead of global option

nak3:
This patch changes to use `spec.HTTPOption` in KIngress instead of
global `httpProtocol` for setting HTTP behavior.

Part of https://github.com/knative/networking/issues/302

Note this patch does not change this TODO comment as it needs a tons of unit test change.

https://github.com/knative-sandbox/net-istio/blob/82f58eb4fb0f8fa44d2b59065058d0adf874a134/pkg/reconciler/ingress/ingress.go#L159-L162

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -151,7 +151,7 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 		// same wildcard host. We need to handle wildcard certificate specially because Istio does
 		// not fully support multiple TLS Servers (or Gateways) share the same certificate.
 		// https://istio.io/docs/ops/common-problems/network-issues/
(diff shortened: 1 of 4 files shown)


nak3:
/hold

Ingress created by DomainMapping still does not set the option. Fixing it by https://github.com/knative/serving/pull/11399.

ZhiminXiang:
Is this ready for review? The PR https://github.com/knative/serving/pull/11399 was merged.

nak3:
No, unfortunately we discovered that https://github.com/knative-sandbox/net-istio/issues/637 should be fixed before this.

Here is a short discussion https://github.com/knative/networking/pull/415#discussion_r639692013 
In short, if one KIngress has different HTTPOption (due to accidentally, concurrency... etc), net-istio controller creates the wrong HTTP behavior.

nak3:
Closing as https://github.com/knative-sandbox/net-istio/pull/658 will fix it.

PR #660, opened 2021-06-03
No need to create HTTP server for wildcard gateway

ZhiminXiang:
For the wildcard Gateway case, a VirtualService ties to both the wildcard Gateway as well as the default global Gateway. (see https://github.com/knative-sandbox/net-istio/blob/main/pkg/reconciler/ingress/ingress.go#L115)

When auto TLS is enabled, the HTTP server of the global Gateway is reconciled to desired shape. So there is no need to add the desired HTTP server to the wildcard Gateway.

/cc @nak3


nak3 (approved):
/lgtm

PR #671, opened 2021-06-11
Update comment about Gateway in configmap

nak3:
The current comment explains about v0.3 which is too old.


markusthoemmes (approved):
/lgtm
/approve

PR #676, opened 2021-06-14
Do not patch local-gateway.mesh to config-istio

nak3:
local-gateway.mesh does not pass some tests so we stop
enabling it during the test.

/kind cleanup


nak3:
/test pull-knative-sandbox-net-istio-integration-tests

`test/conformance: TestIngressConformance/8/visibility/split` failed. I think flake.

ZhiminXiang (approved):
/lgtm

PR #682, opened 2021-06-18
Use terminationDrainDuration in MeshConfig.defaultConfig

nak3:
Istio upstream drops `TERMINATION_DRAIN_DURATION_SECONDS` since 1.10
https://github.com/istio/istio/commit/f6502508b04c38f9678c300821dd54e439b832ed.

/kind cleanup

### DIFF
--- a/third_party/istio-head/istio-ci-mesh.yaml
+++ b/third_party/istio-head/istio-ci-mesh.yaml
@@ -30,17 +30,13 @@ spec:
 
   meshConfig:
     defaultConfig:
-      proxyMetadata:
-        TERMINATION_DRAIN_DURATION_SECONDS: "20"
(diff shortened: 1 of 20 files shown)


nak3:
Current latest-mesh test always fails with TestDestroyPod* as https://testgrid.k8s.io/r/knative-own-testgrid/serving#istio-latest-mesh this patch should fix it.

julz:
Would this change mean we don't drain properly on 1.9.* or does the new config work there too?

nak3:
Thank you! The new config works on 1.9.* (istio-stable) so applied it on 1.9 as well.

ZhiminXiang (approved):
/lgtm

PR #686, opened 2021-06-22
fix webhook name

dprotaso:


dprotaso:
/assign @julz @ZhiminXiang 

This someone didn't get caught by the presubmit which was surprising

julz (approved):
/lgtm
/approve

dprotaso:
> This someone didn't get caught by the presubmit which was surprising

I realize the webhook operates on knative serving types and not networking types - hence why this didn't fail in the presubmit checks

PR #693, opened 2021-06-28
fix ordering of resources to ensure webhook config is last

dprotaso:
This resolves a race when the webhook configuration is applied _before_ the webhook deployment/service is applied


dprotaso:
/hold cancel

verified locally

julz (approved):
/lgtm
/approve

PR #708, opened 2021-07-20
[main] Upgrade to latest dependencies 

nak3:
This patch contains https://github.com/knative-sandbox/net-istio/pull/707 with the following fix in third_party/library.sh.

The `yq eval` with `select` produces `{}` when it runs agianst empty file.

```
$ touch a.yaml
$ yq eval "select(.kind != \"test1\")" ./a.yaml
{}
```

So this patch changes the script to break the loop when `${tmpfile}` is empty.

/cc @dprotaso @ZhiminXiang

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,6 @@ require (
 	k8s.io/apimachinery v0.20.7
 	k8s.io/client-go v0.20.7
 	knative.dev/hack v0.0.0-20210622141627-e28525d8d260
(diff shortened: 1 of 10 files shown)


nak3:
And the usage of Client needs to be updated like https://github.com/knative/networking/commit/bf0fc8442924badb4fb87572d23f5025887a1925

ZhiminXiang (approved):
/lgtm

PR #747, opened 2021-09-03
drop scheduled actions these fill up the job queue causing delays

dprotaso:


.github/workflows/kind-e2e.yaml
```diff
@@ -7,9 +7,6 @@ on:
   push:
     branches: [ 'main', 'release-*' ]
 
-  schedule:
-  - cron: '0 * * * *'
```
nak3, line 11: Is it better to change to run everyday like `'0 0 * * *'`?

dprotaso: I'm a bit mixed because

1) It's not clear folks are looking at the failures of these jobs when it's spamming slack
2) There exist continuous testing in prow - which doesn't affect the limit Github Action limit
3) It's hard to jitter all the e2e tests runs across all repos - ie. if they all run at `0 0...` then it'll create a big queue at that time

nak3: I understand. That makes sense. Thank you.

nak3:
/assign @ZhiminXiang 

LGTM

ZhiminXiang (approved):
/lgtm

PR #781, opened 2021-10-08
Use GKE 1.20 for CI

nak3:
Current CI is very unstable as https://testgrid.k8s.io/r/knative-own-testgrid/net-istio#latest
This patch changes to use GKE 1.20.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -18,7 +18,7 @@ source $(dirname $0)/e2e-common.sh
 
 
 # Script entry point.
-initialize $@  --skip-istio-addon
+initialize $@  --skip-istio-addon --cluster-version=1.20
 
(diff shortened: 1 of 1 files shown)


nak3:
/hold

I wonder that this GKE version should pin to hack repo https://github.com/knative/hack/blob/815cd312d65c62d69fd9cd9fa3819e867c4c2b9b/test/e2e-tests.sh as all repo seems to have the performance issue.

nak3:
/cc @ZhiminXiang @dprotaso 

This makes CI more stable. I think we should use GKE 1.20 until https://github.com/knative/serving/issues/12091 was solved. Some PRs are blocked due to CI issue such as https://github.com/knative-sandbox/net-istio/pull/780

PR #787, opened 2021-10-16
fix go1.17 related issues

dprotaso:
This fixes:

1) Image uploading 
  _We shouldn't be explicitly marking module names with the `vendor` prefix - this should just work with newer versions of go_
2) KinD e2e tests to use go modules (so upload is the same in actions & prow)


dprotaso:
ouch - API server is off

> error: failed to create clusterrolebinding: Post "https://34.74.119.103/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?fieldManager=kubectl-create": net/http: TLS handshake timeout

nak3 (approved):

PR #804, opened 2021-11-05
Add k8s recommended labels

psschwei:
# Changes

Adds Kubernetes recommended labels to networking as discussed in https://github.com/knative/networking/issues/552

Note that changes to the labels in `vendor/knative.dev/networking` should be handled by the automation once https://github.com/knative/networking/pull/555 merges. 

/kind enhancement

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -18,8 +18,8 @@ metadata:
   # These are the permissions needed by the Istio Ingress implementation.
   name: knative-serving-istio
   labels:
-    app.kubernetes.io/name: istio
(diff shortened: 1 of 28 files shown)


dprotaso (changes requested):
likewise - call it `net-istio` vs. `istio`

dprotaso:
@nak3 I don't see anything in the PR that would start causing things to fail.

The one difference to when an earlier revision was passing was it used GKE 1.20 - but all the new failures are GKE 1.21

nak3:
/retest

Yeah :cry: The CI is unstable but I believe infra side issue.

Issue #719, opened 2021-08-05
`ListProbeTargets()` does not list probe targets accurately


nak3:
Current [`ListProbeTargets()`](https://github.com/knative-sandbox/net-istio/blob/d7625fa51eb7e53757e4af95bc09ac86f07c921a/pkg/reconciler/ingress/lister.go#L61) does:

1. List Istio Gateways specified in config-istio by `qualifiedGatewayNamesFromContext()`
2. List the ingress service by the Gateway selector.
3. Add the schema (`http` or `https`) defined in the Istio Gateway listed by step-1 to the probe targets.

However, when autoTLS is enabled, Istio Gateways are automatically created by net-istio. So the above process (especially the combination of step-1 & step-3) does not work well.

Step-3 should determine the protocol by `ingress.spec.httpoption`, but net-istio supports manual TLS certs [1] and it will not work in that case.

[1] https://knative.dev/docs/serving/using-a-tls-cert/#manually-adding-a-tls-certificate

nak3:
The another failure scenario is discussed in the slack https://knative.slack.com/archives/C0186KU7STW/p1637206270281700?thread_ts=1637171885.256400&cid=C0186KU7STW

If an Istio Gateway (`knative-ingress-gateway`) has `tls.mode` without `MUTUAL` the prober recoginize the target as TLS and it  tries to probe with `https` scheme.
For example, the following gateway fails. 
```
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - example.com
    port:
      name: http-app-http
      number: 80
      protocol: HTTP
    tls:
      httpsRedirect: true
  - hosts:
    - example.com
    port:
      name: https-app-https
      number: 443
      protocol: HTTPS
    tls:
      credentialName: app-cert
      mode: MUTUAL
  - hosts:
    - example.com
    port:
      name: https-app-https-simple
      number: 443
      protocol: HTTPS
    tls:
      credentialName: app-cert
      mode: SIMPLE
```

PR #814, opened 2021-11-18
Exclude skipped targets/hosts from ListProbeTargets

psaia:
# Changes

Exclude the skipped hosts/servers (`MUTUAL`, redirects, etc) from the list of probe targets returned from `ListProbeTargets` so kn doesn't attempt to probe an unresolvable server. The specific side effect/bug that this fixes is described in the following places:

* [GH Issue comment](https://github.com/knative-sandbox/net-istio/issues/719#issuecomment-972834067) - I have not confirmed that this fixes the problem outlined by @nak3 in the first comment, only the second.
* [Slack conversation](https://knative.slack.com/archives/C0186KU7STW/p1637171885256400)

/kind bug

### DIFF
--- a/pkg/reconciler/ingress/lister.go
+++ b/pkg/reconciler/ingress/lister.go
@@ -73,10 +73,11 @@ func (l *gatewayPodTargetLister) ListProbeTargets(ctx context.Context, ing *v1al
 		if err != nil {
 			return nil, fmt.Errorf("failed to get Gateway %q: %w", gatewayName, err)
 		}
-		targets, err := l.listGatewayTargets(gateway)
+		targets, excludedHosts, err := l.listGatewayTargets(gateway)
 		if err != nil {
 			return nil, fmt.Errorf("failed to list the probing URLs of Gateway %q: %w", gatewayName, err)
 		}
+
 		if len(targets) == 0 {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


psaia:
I signed the CLA

/retest

psaia:
I've signed the CLA.

nak3:
> GH Issue comment - I have not confirmed that this fixes the problem outlined by @nak3 in the first comment, only the second.

Yeah, the first issue in https://github.com/knative-sandbox/net-istio/issues/719 is my concern. autoTLS should not be related to this change but it will be effected due to the bug :sweat:

psaia:
> > GH Issue comment - I have not confirmed that this fixes the problem outlined by @nak3 in the first comment, only the second.
> 
> Yeah, the first issue in #719 is my concern. autoTLS should not be related to this change but it will be effected due to the bug 😓

By that, do you mean that this change may cause an issue with autoTLS?

pkg/reconciler/ingress/lister.go
```diff
@@ -73,23 +73,35 @@ func (l *gatewayPodTargetLister) ListProbeTargets(ctx context.Context, ing *v1al
...
 			continue
 		}
 		for _, target := range targets {
+			// Pick a single host since they all end up being used in the same
+			// VirtualService and will be applied atomically by Istio. However,
+			// it's important to choose a hostname which isn't excluded from the
+			// initial list of gateway targets.
+			var host string
+			for _, hostName := range hostsByGateway[gatewayName].List() {
+				if !excludedHosts.Has(hostName) {
```
nak3, line 91: If `myapp.example.com` has `tls.HttpsRedirect`, prober will exclude the host from target even with `HTTPS`?

nak3:
Yes, the issue may not be serious as it will exclude the target from the prober unexpectedly, though. So, I think it is risky to fix the second bug in https://github.com/knative-sandbox/net-istio/issues/719 without the first one fixed.

psaia:
@nak3, ok. I've updated the logic to not exclude redirects. Understood about the risk. I'd try to fix the first issue in order to get this one out the door, but I don't fully understand the problem as I'm not familiar with the mechanics of autoTLS.

pkg/reconciler/ingress/lister.go
psaia: Good point. I've removed the redirected hosts from the list of excluded hosts.

Issue #585, opened 2021-04-06
enabling virtualservice status breaks update tests


dprotaso:
I've observed the virtual service `status.observedGeneration` doesn't change so the net-istio reconciler thinks it's stale and doesn't update. I'm going to disable this flag for the latests tests for now

> I realize the `config-istio` get's cleared when we `ko`'d the config

So because of this we never really ran e2e tests with the virtual status enabled - thus I think this is a legit bug.

_Originally posted by @dprotaso in https://github.com/knative-sandbox/net-istio/issues/583#issuecomment-814036967_

dprotaso:
I created an upstream issue: https://github.com/istio/istio/issues/31931

dprotaso:
/good-first-issue

We should remove this functionality from net-istio

PR #818, opened 2021-11-25
Disable enable-virtualservice-status on istio-head test

nak3:
Istio head test keeps failing as https://testgrid.k8s.io/r/knative-own-testgrid/serving#istio-head-no-mesh
This is most probably due to `enable-virtualservice-status: true` and nobody takes care of the result.

This patch changes:
- to disables enable-virtualservice-status until https://github.com/knative-sandbox/net-istio/issues/585 was fixed.
- to update Istio version as Stable=1.11.4 and Latest=1.12.0.

/cc @dprotaso @ZhiminXiang


dprotaso:
/lgtm
/approve

FYI -  if you want to keep serving e2e pinned to istio 1.11 you'll want to update
https://github.com/knative/serving/blob/abf4d6933cd03187e02b41bb20e2ae5533de8ee2/test/e2e-common.sh#L25 to be stable

nak3:
Thank you! I think 1.12 should be fine but will pin to 1.11 if it fails.

Issue #782, opened 2021-10-09
Drop Istio stable (or latest) from test


nak3:
Currently Istio has __three__ version tests as:

head: https://testgrid.k8s.io/r/knative-own-testgrid/serving#istio-head-no-mesh
latest: https://testgrid.k8s.io/r/knative-own-testgrid/serving#istio-latest-no-mesh
stable: https://testgrid.k8s.io/r/knative-own-testgrid/serving#istio-stable-no-mesh

We should drop the latest or stable from the CI for the following reasons:
- The `stable` is an old version. Istio community does not support it with the latest k8s version so we say stable but it is sometimes not supported by Istio community (Please also see https://github.com/knative/serving/issues/12064#issuecomment-931727574)
- We needed stable and latest before because Istio had a big gap between the minor versions but they don't have the gap recently.
- Nobody takes care of `head`. We should take care `head` and the latest (or stable) version only.

PR #820, opened 2021-11-27
Remove istio-stable from test

nak3:
https://github.com/knative-sandbox/net-istio/issues/782 described several reasons to remove `istio-stable`.
In short, we should maintain `head` and `latest` rather than adding `stable` which makes complication.

Fix https://github.com/knative-sandbox/net-istio/issues/782


nak3:
Depends on https://github.com/knative/test-infra/pull/2951

PR #826, opened 2021-12-06
Remove istio-kind-mesh.yaml

nak3:
This patch removes istio-kind-mesh.
Kind test always runs no mesh and the istio-kind-mesh is
actually wrong and no different from istio-kind-no-mesh.

/cc @dprotaso @ZhiminXiang


third_party/istio-head/generate-manifests.sh
```diff
@@ -16,6 +16,4 @@
 
 source "$(dirname $0)/../library.sh"
 
-generate HEAD "$(dirname $0)" \
-      --set values.pilot.env.PILOT_ENABLE_STATUS=true \
```
nak3, line 20: I missed to delete this in https://github.com/knative-sandbox/net-istio/pull/818
If I should send a separated PR, I do.

This script is not used anyway https://github.com/knative-sandbox/net-istio/blob/f92596ff8ae2b2a2b37def30944bb234561c91b4/hack/update-codegen.sh#L64-L66

dprotaso (approved):
/lgtm
/approve

PR #846, opened 2022-01-12
Use different Lister between autoTLS enabled or disabled

nak3:
As https://github.com/knative-sandbox/net-istio/issues/719 described, Istio Gateways are auto generated when autoTLS is enabled.
However, current ListProbeTargets lists the probe targets based on the single Gateway which is used by autoTLS disabled.

Hence this patch creates another listProbetargets function to list the target when autoTLS is enabled.

Fix https://github.com/knative-sandbox/net-istio/issues/719


nak3:
/hold
I haven't added the unit tests yet. I will add the unit tests if somebody could review this PR.

PR #874, opened 2022-02-14
Add the EnvoyFilter to support websocket for DomainMapping

ZhiminXiang:
# Changes

- Add the EnvoyFilter to support websocket for DomainMapping

For more details, see https://github.com/knative/serving/issues/12601#issuecomment-1032017576

**Release Note**

```release-note
Fix the bug to support custom domain Websocket for Istio ingress implementation.
```

### DIFF
--- a/config/700-websocket-envoyfilter.yaml
+++ b/config/700-websocket-envoyfilter.yaml
@@ -1,0 +1,42 @@
+# Copyright 2022 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This Envoyfilter is needed to support websocket for DomainMapping.
+apiVersion: networking.istio.io/v1alpha3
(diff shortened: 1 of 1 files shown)


config/700-websocket-envoyfilter.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+# This Envoyfilter is needed to support websocket for DomainMapping.
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: dm-websocket-filter
+  namespace: istio-system
+  labels:
+    app.kubernetes.io/component: net-istio
+    app.kubernetes.io/name: knative-serving
```
nak3, line 23: Should we add the same labels with other configuration like https://github.com/knative-sandbox/net-istio/blob/d17e236d799a12687a4bb29b0d9d1a35592c35d0/config/400-config-istio.yaml#L21-L25 

AFAIK, knative/operator uses `networking.knative.dev/ingress-provider` so we need it at least.

config/700-websocket-envoyfilter.yaml
```diff
@@ -0,0 +1,42 @@
...
+spec:
+  workloadSelector:
+    labels:
+      app: istio-ingressgateway
```
nak3, line 27: I think this will not work when we use another ingress-gateway instead of the default `istio-ingressgateway`.
Do you have any good idea how we can take care of it? Or we will just document it?

ZhiminXiang: That's true. Currently we already hard coded the selector though in our Gateway https://github.com/knative-sandbox/net-istio/blob/main/config/203-local-gateway.yaml#L31 and https://github.com/knative-sandbox/net-istio/blob/main/config/202-gateway.yaml#L29. I am curious what mechanism we use for those Gateways?

Issue #993, opened 2022-10-07
Enable k8s 1.25 test


nak3:
Due to https://github.com/istio/istio/issues/41220,  `istioctl manifest generates` does not produce the compatible version for k8s 1.25.

We disabled 1.25 temporary but we should figure out this somehow - e.g. use `istioctl install` for the test.

https://github.com/knative-sandbox/net-istio/blob/f079fd3dec63d08d6a4be4711e7e947e5b7fd129/.github/workflows/kind-e2e.yaml#L25-L28

PR #994, opened 2022-10-07
Add workaround for k8s 1.25

nak3:
Due to an issue in `istioctl manifest generates`, the manifest does not  produce `policy/v1`.

This patch adds the workaround which replaces the version by sed in script.

Fix https://github.com/knative-sandbox/net-istio/issues/993


psschwei (commented):
/lgtm

PR #1012, opened 2022-11-09
Make the webhook port number configurable

sel:
Make the webhook port number configurable

Previously the webhook listened on a fixed port, 8443, which can clash
with other services when the webhook is run on the host network in
Kubernetes, which is required when using some CNI implementations,
notably Calico on EKS [1].

Enable configuration of the webhook listen port through the environment
variable WEBHOOK_PORT by using the existing API from the KNative SDK.

[1] https://projectcalico.docs.tigera.io/getting-started/kubernetes/managed-public-cloud/eks#install-eks-with-calico-networking

Signed-off-by: Steve Larkin <steve.larkin@gmail.com>

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -81,7 +81,7 @@ func main() {
 	ctx := webhook.WithOptions(signals.NewContext(), webhook.Options{
 		ServiceName: "net-istio-webhook",
 		SecretName:  "net-istio-webhook-certs",
-		Port:        8443,
+		Port:        webhook.PortFromEnv(8443),
 	})
 
 	sharedmain.WebhookMainWithContext(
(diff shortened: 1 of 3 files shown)


nak3:
LGTM.

I think we should align with serving's change like https://github.com/knative/serving/pull/7248 so I would add `WEBHOOK_PORT` in [config/500-webhook-deployment.yaml](https://github.com/knative-sandbox/net-istio/blob/main/config/500-webhook-deployment.yaml).

sel:
> add `WEBHOOK_PORT` in [config/500-webhook-deployment.yaml](https://github.com/knative-sandbox/net-istio/blob/main/config/500-webhook-deployment.yaml).

Thanks @nak3! I have made this change in 017fbfc. In that change I also modified `config/500-webhook-service.yaml` to use named targetPorts to avoid needing to patch that file when using a non-default webhook port.

nak3:
/lgtm
/assign @AngeloDanducci 

>  In that change I also modified config/500-webhook-service.yaml to use named targetPorts to avoid needing to patch that file when using a non-default webhook port.

Oh, thank you so much!

AngeloDanducci (approved):

PR #1028, opened 2022-11-24
add config network as a volume mount

AngeloDanducci:
# Changes

- adds config-network as a configmap for the net-istio controller, enables proxy protocol probes

/kind enhancement

Needs the https://github.com/knative/networking/pull/712 PR first

### DIFF
--- a/config/500-controller.yaml
+++ b/config/500-controller.yaml
@@ -87,5 +87,13 @@ spec:
         - name: profiling
           containerPort: 8008
 
+        volumeMounts:
+          - name: config-network
+            mountPath: /etc/config-network
+            readOnly: true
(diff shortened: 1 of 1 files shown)


nak3:
Is it not OK to get the value via API like this instead of mounting the file?

https://github.com/knative-sandbox/net-istio/blob/c5d5ea57dc2ca280d48ce243122bca22915c7624/pkg/reconciler/serverlessservice/serverlessservice.go#L51

dprotaso:
@nak3 is correct - the controllers have a config 'store' that register the config maps that want to watch

ReToCode:
Please re-open if this is still needed/active.

/close

PR #1030, opened 2022-11-29
Explicitly set minProtocolVersion

nak3:
As per https://github.com/knative/serving/issues/13451, Istio does not set 
`minProtocolVersion` to TLS v1.2.

This patch sets it explicitly as TLS v1.1 is EOL now.


nak3:
/hold

Need to add unit test.

Depends on the reply of https://github.com/knative/serving/issues/13451#issuecomment-1329903540
If the issue happens only on the old Istio versions, we should rely on the Istio's default behavior.

nak3:
Closing.
As per https://istio.io/latest/docs/reference/config/networking/gateway/, the default is TLS v1.2.

milank78git:
Hi @nak3 , the Istio verion in our company is 1.13.3 and 1.15  . I am not able to upgrade it easily.

The defaut TLS verison is not set in this Istio version, therefore I can not rely on the default being TLS_v1.2. https://istio.io/v1.15/docs/reference/config/networking/gateway https://istio.io/v1.13/docs/reference/config/networking/gateway

psschwei (commented):
/lgtm

evankanderson (commented):
/lgtm
/approve

PR #1043, opened 2022-12-29
Ensure that a Secret label value does not exceed 63 length

SaschaSchwarze0:
# Changes

I have Knative and Istio together with net-istio running, and I created a domain mapping with a referenced TLS secret that has a name that exceeds 63 characters. Knative Net-Istio tries to synchronize this secret to the istio-system namespace and fails because it uses the secret name as label value:

```
Failed to create Secret istio-system/areallyverylongdomainnamethatexcd8923dee789a086a0ac46d3046cbec7-9c37460d-0d61-4f36-a65a-398b49d40f48: Secret "areallyverylongdomainnamethatexcd8923dee789a086a0ac46d3046cbec7-9c37460d-0d61-4f36-a65a-398b49d40f48" is invalid: metadata.labels: Invalid value: "areallyverylongdomainnamethatexceeds63charactersintotal.somethinglong.com": must be no more than 63 characters
```

I am changing the logic depending on the length of the secret name:

* Nothing changes for secrets with a name length up to 63 characters. This ensures that all existing synchronized secrets can still be matched to the origin secret.
* For a secret with a name length larger than 63 characters, I am putting the secret name into an annotation because there is [logic that must identify the origin secret from the copy](https://github.com/knative-sandbox/net-istio/blob/release-1.8/pkg/reconciler/ingress/ingress.go#L263) and therefore needs the secret name. I am putting in a 63 characters length label value using a hash. This is necessary for the deletion case where it must [find the copy in the istio-system namespace based on a label selector](https://github.com/knative-sandbox/net-istio/blob/release-1.8/pkg/reconciler/ingress/ingress.go#L391-L392).

/kind bug

**Release Note**

```release-note
You can now reference secrets with a name that has a length longer than 63 characters in a domain mapping
```

**Docs**

N/A


AngeloDanducci:
Looks like a fix to a bug we didn't know we had yet, thanks!

cc @nak3

evankanderson (commented):
Unless we're listing secrets by label somewhere else, it seems like these probably should have been annotations to begin with.  However, for a bug fix, I'm 👍  on doing the smallest surgical change, so

/lgtm
/approve

PR #1053, opened 2023-02-04
Drop Istio HEAD

dprotaso:
We don't really test istio@HEAD anywhere in CI - not sure if it's worth it anymore.


nak3:
/cc @AngeloDanducci @psschwei 

LGTM

psschwei (commented):
/lgtm

PR #1057, opened 2023-02-09
Ensure wildcard certificate secrets can also track back their source secret

SaschaSchwarze0:
# Changes

When recently doing https://github.com/knative-sandbox/net-istio/pull/1043, I was already surprised that a wildcard certificate copy in the Istio system namespace is NOT getting any labels to track back the original secret.

I now verified that this is indeed causing issues:

1. An update of the original secret is not propagated to the copy.
2. The secret copy is never deleted even if it is not needed anymore.

I am therefore setting the same labels and annotation on a wildcard secret in this PR. The main change is in pkg/reconciler/ingress/resources/secret.go.

The purpose of the change in pkg/reconciler/accessor/core/secret.go is to fix all existing secret copies' labels. The test case change in pkg/reconciler/ingress/ingress_test.go is related to this given the missing label is now also updated.

/kind bug

**Release Note**

```release-note
Updates to secrets with wildcard certificate are now properly propagated to the istio-system namespace
```

**Docs**

N/A

### DIFF
--- a/pkg/reconciler/accessor/core/secret.go
+++ b/pkg/reconciler/accessor/core/secret.go
@@ -59,10 +59,11 @@ func ReconcileSecret(ctx context.Context, owner kmeta.Accessor, desired *corev1.
 		return nil, kaccessor.NewAccessorError(
 			fmt.Errorf("owner: %s with Type %T does not own Secret: %s", owner.GetName(), owner, secret.Name),
 			kaccessor.NotOwnResource)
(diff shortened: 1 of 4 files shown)


evankanderson (commented):
/lgtm
/approve

We should make sure this works with https://github.com/knative-sandbox/net-istio/pull/920, which might have affected the existing behaviors.

PR #1061, opened 2023-02-15
bump istio to 1.17.0

dprotaso:
### Release Note

```release-note
Bump Istio to v1.17.0
```


go.mod
```diff
@@ -2,17 +2,24 @@ module knative.dev/net-istio
...
 go 1.18
 
+replace (
+	k8s.io/api => k8s.io/api v0.25.6
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.25.6
+	k8s.io/apimachinery => k8s.io/apimachinery v0.25.6
+	k8s.io/client-go => k8s.io/client-go v0.25.6
+	k8s.io/code-generator => k8s.io/code-generator v0.25.6
+)
+
```
dprotaso, line 12: pinned because istio is pulling in l8s 0.26.0 deps - but we're at 0.25

dprotaso:
note: `third_party/go.mod` is essentially like a `go` `.gitignore`

see: https://github.com/golang/go/wiki/Modules#can-an-additional-gomod-exclude-unnecessary-content-do-modules-have-the-equivalent-of-a-gitignore-file

PR #1085, opened 2023-03-28
Add support for `system-internal-tls` in net-istio

### LINKED ISSUE #1063: Support system-internal-tls in net-istio
Larger description in the [Feature Track document](https://docs.google.com/document/d/1XE7UzgQlVVtAb7ULSqOyKCaIHtm8zMF35ainp1JmwyY/edit?resourcekey=0-e1MXRxQaalq-EHW46ZCCWg#heading=h.n8a530nnrb)
Parent-issue: https://github.com/knative/serving/issues/11906

**Summary**
net-istio should support calling activator / backends with a known CA key and subject name.

/kind feature-request

ReToCode:
# Changes
- Adds support for `system-internal-tls` in `net-istio`
- Adds a build matrix for conformance and e2e tests with upstream TLS enabled
- Verified on Serving: https://github.com/knative/serving/pull/14494

# Things to discuss
1. `config/700-istio-secret.yaml` is necessary because the CA secrets needs to be in `istio-system` namespace for istio to pick it up. It's currently the only way to get the CA there. This is why https://docs.google.com/document/d/1YdcdBVg_zpT4WSNRsWlihut5JuLddOTIggFvLni3A28/edit?disco=AAAAtaDADjo proposes to add the CA to `KIngress`
2. Handling of DomainMappings is a bit hacky. It is adapted based on `net-kourier`. If we do https://github.com/knative/serving/issues/13472 this handling can be dropped (here and in `net-kourier` and `net-contour`)
3. Handling/detection of `http2` is currently based on service naming, which is not ideal. Thus https://docs.google.com/document/d/1YdcdBVg_zpT4WSNRsWlihut5JuLddOTIggFvLni3A28/edit?disco=AAAAtaDADjo also proposes to add the upstream protocol to the `KIngress` resource to avoid this in the future.

**Release Note**
No release notes yet, as this is still an internal/alpha feature.

**Docs**
No docs yet, as this still an internal/alpha feature.

/kind enhancement
Fixes #1063


evankanderson:
I like the suggestion to do https://github.com/knative/serving/issue/13472, particularly since we've seen it show up three times and within-cluster TLS is not an unreasonable desire.

(Non-blocking, will review the rest soon.)

evankanderson:
One other thought: with websockets being defined for http2, I wonder whether we should simply run http2 to the queue-proxy and do the down-conversion there.

Downside: we may get scale-from-zero wakeups for requests the app won't actually be able to handle.

Upside: people can stop needing to worry about naming their port.

pkg/reconciler/ingress/resources/destinationrule.go
```diff
@@ -0,0 +1,64 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/networking/pkg/apis/networking/v1alpha1"
+	"knative.dev/pkg/kmeta"
+)
+
+const (
+	// has to match config/700-istio-secret.yaml
+	knativeServingCertsSecret = "knative-serving-certs"
+
+	// has to match https://github.com/knative-sandbox/control-protocol/blob/main/pkg/certificates/constants.go#L21
```
evankanderson, line 31: I think David is making this a public constant shortly.

.github/workflows/kind-e2e.yaml
```diff
@@ -164,6 +189,12 @@ jobs:
            --ingressendpoint="${IPS[0]}" \
            --ingressClass=istio.ingress.networking.knative.dev \
            --cluster-suffix=$CLUSTER_SUFFIX
+        
+        echo "# DEBUG"
```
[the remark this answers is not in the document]
ReToCode: No I'm trying to find out why the builds fail. :-)

pkg/reconciler/ingress/ingress.go
```diff
@@ -188,6 +191,14 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	}
 	gatewayNames[v1alpha1.IngressVisibilityExternalIP].Insert(resources.GetQualifiedGatewayNames(ingressGateways)...)
 
+	if config.FromContext(ctx).Network.InternalEncryption {
+		logger.Info("Creating/Updating DestinationRules for internal-encryption")
```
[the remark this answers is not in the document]
ReToCode: I just aligned with the `logger.Info("Creating/Updating VirtualServices")` below. I would propose to let it stay at info.

pkg/reconciler/ingress/resources/destinationrule.go
ReToCode: Will use that as soon as available (in this PR or a followup depending on merge order).

ReToCode:
> I like the suggestion to do https://github.com/knative/serving/issue/13472, particularly since we've seen it show up three times and within-cluster TLS is not an unreasonable desire.

I synced with @KauzClay. Well come up with some GH-issues to track tasks out of the Findings document shortly to start working on that. But we'd need to agree on the contract and `KIngress` changes first.

> One other thought: with websockets being defined for http2, I wonder whether we should simply run http2 to the queue-proxy and do the down-conversion there.

I think it would be worth to look into this. Should we track/discuss this in a new issue?

> Initial review; this was less painful than I anticipated.

That's a hopefully a good thing? ;)

pkg/reconciler/ingress/resources/destinationrule.go
```diff
@@ -0,0 +1,64 @@
...
+	knativeFakeDnsName = "data-plane.knative.dev"
+)
+
+// MakeInternalEncryptionDestinationRule creates a DestinationRule that enables upstream TLS
+// on for the specified host
+func MakeInternalEncryptionDestinationRule(host string, ing *v1alpha1.Ingress, http2 bool) *v1alpha3.DestinationRule {
+	dr := &v1alpha3.DestinationRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            host,
+			Namespace:       ing.Namespace,
```
nak3, line 41: I think we should propagate label and annotation from the Ingress.

nak3 (commented):
I guess we should add the same https://github.com/knative-sandbox/net-istio/pull/114 for DestinationRules in `cmd/controller/main.go`?

pkg/reconciler/ingress/resources/destinationrule.go
ReToCode: Done.

ReToCode:
> I guess we should add the same https://github.com/knative-sandbox/net-istio/pull/114 for DestinationRules in cmd/controller/main.go?

@nak3 any idea why we are configuring `v1beta1` [here](https://github.com/knative-sandbox/net-istio/pull/1085/files#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554dR35) but are using `v1alpha3` everywhere else?

pkg/reconciler/ingress/resources/destinationrule.go
ReToCode: We don't have a dependency on `control-protocol` here and introducing it just for the constant is a bit much. I heard that we maybe are  going to refactor the certificate part out of `control-protocol` into `networking`. If that is so, we can keep it and use it from `networking` when it is done?

nak3:
I don't know the reason. As far as I researched the commit history (you might already know but...)
* https://github.com/knative-sandbox/net-istio/pull/147 bumped to `v1beta1`
* https://github.com/knative-sandbox/net-istio/pull/203 reverted to `v1alpha3` but missed `cmd/controller/main.go` (not sure if it is accident or intention though)

The revert is waiting for https://github.com/knative/serving/issues/8896 but https://github.com/knative/serving/issues/8896 should be fixed most probably so we should bump the API to `v1beta1`. I opened https://github.com/knative-sandbox/net-istio/issues/1087

pkg/reconciler/ingress/resources/destinationrule.go
nak3: I think it is better to use the `control-protocol` even adding the deps. Otherwise we have to keep in mind of this place.

nak3:
Overall looks good to me.

With said, Istio has a large variety of network configuration so  I am worried that adding a non-edge network feature conflicts with their features :disappointed_relieved:  For example, does this `internal-encryption` in `net-istio` assume that there is no sidecar at all? Or it can work with the partial mesh (moreover using PERMISSIVE mode but it means HTTPS over mTLS)?...etc
This is still alpha feature so we can move forward, though.

nak3:
@evankanderson Would you like to have another review?

ReToCode:
Rebased and istio version bumped.

@evankanderson any thoughts about https://github.com/knative-sandbox/net-istio/pull/1085#issuecomment-1495211120 and https://github.com/knative-sandbox/net-istio/pull/1085#discussion_r1156603079? For the latter, what is the plan to move/migrate stuff from `control-protocol` to `networking`? Can we maybe start with migrating the constants instead of adding another dependency on `control-protocol`?

ReToCode:
@nak3 any idea why this fails just with ambient?

>     --- FAIL: TestIstioProbing/HTTP (9.69s)
        probing_test.go:235: [istio-probing-http-pkekzbpr] Using port 61
        probing_test.go:235: [istio-probing-http-pkekzbpr] Using containerPort 8050
        util.go:877: ingressendpoint: "172.18.0.3"
        spoof.go:110: Spoofing istio-probing-http-pkekzbpr.example.com -> 172.18.0.3
        probing_test.go:289: Failed to probe the Service: failed to probe http://istio-probing-http-pkekzbpr.example.com/: response: status: 400, body: Client sent an HTTP request to an HTTPS server.

AFAIK, this test calls a created service via Istio (via spoof client). With `system-internal-tls` enabled, this should not be a different call for the client (the test in this case). It should be able to still call istio with http/https (depending on the test case) and istio should internally call all the target services using TLS.

ReToCode:
Also, back to the original discussion, we probably only want to enable this in `no-mesh` mode anyways. As it makes no sense when mesh or ambient is used anyways. So maybe we should just test this in `no-mesh` mode? @nak3  WDTY?

nak3:
Yes, I think we should just test this in no-mesh mode.

ReToCode:
@nak3 this is ready now. I verified on Serving [here](https://github.com/knative/serving/pull/14494). Can you take another look?

pkg/reconciler/ingress/resources/destinationrule.go
```diff
@@ -0,0 +1,76 @@
...
+				},
+			},
+		},
+	}
+
+	// Populate the Ingress labels.
+	dr.Labels = kmap.Filter(ing.GetLabels(), func(k string) bool {
+		return k != RouteLabelKey && k != RouteNamespaceLabelKey
+	})
+	dr.Labels[networking.IngressLabelKey] = ing.Name
```
nak3, line 66: I think it is alright to add ingress label into DestinationLabel without this filter. But it was added to be consistent with VirtualService, wasn't it? 
Either is fine so let's keep as it is.

nak3 (commented):
LGTM! Only minor comments.

pkg/reconciler/ingress/resources/destinationrule.go
ReToCode: > But it was added to be consistent with VirtualService, wasn't it?

Yes, exactly.

ReToCode:
@dprotaso do you want to take a look as well?

PR #1095, opened 2023-04-13
Add missing probes

skonto:
# Changes
- Add probes to net-istio-controller and net-istio-webhook

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -84,8 +84,9 @@ func main() {
 		Port:        webhook.PortFromEnv(8443),
 	})
 
+	ctx = sharedmain.WithHealthProbesDisabled(ctx)
 	sharedmain.WebhookMainWithContext(
-		sharedmain.WithHealthProbesDisabled(ctx), "net-istio-webhook",
+		ctx, "net-istio-webhook",
 		certificates.NewController,
 		NewDefaultingAdmissionController,
 		NewConfigValidationController,
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -85,7 +85,7 @@ func main() {
 	})
 
 	sharedmain.WebhookMainWithContext(
-		ctx, "net-istio-webhook",
+		sharedmain.WithHealthProbesDisabled(ctx), "net-istio-webhook",
 		certificates.NewController,
 		NewDefaultingAdmissionController,
 		NewConfigValidationController,
--- a/config/500-controller.yaml
+++ b/config/500-controller.yaml
@@ -81,20 +81,20 @@ spec:
           seccompProfile:
             type: RuntimeDefault
 
-        livenessProbe:
-          httpGet:
-            path: /health
-            port: probes
-            scheme: HTTP
-          periodSeconds: 5
-          failureThreshold: 5
         readinessProbe:
           httpGet:
             path: /readiness
             port: probes
             scheme: HTTP
           periodSeconds: 5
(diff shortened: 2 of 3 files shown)


cmd/webhook/main.go
```diff
@@ -85,7 +85,7 @@ func main() {
 	})
 
 	sharedmain.WebhookMainWithContext(
-		ctx, "net-istio-webhook",
+		sharedmain.WithHealthProbesDisabled(ctx), "net-istio-webhook",
```
skonto, line 88: Deactivating the default endpooint as we will use the webhook one.

config/500-webhook-deployment.yaml
```diff
@@ -82,6 +82,23 @@ spec:
...
+              - name: k-kubelet-probe
+                value: "webhook"
+        livenessProbe:
+          periodSeconds: 1
+          httpGet:
+            scheme: HTTPS
+            port: 8443
+            httpHeaders:
+              - name: k-kubelet-probe
+                value: "webhook"
```
nak3, line 100: Should we have same values with serving's webhook (i.e. should not be same between readiness and liveness )? - https://github.com/knative/serving/blob/main/config/core/deployments/webhook.yaml#L122-L123

Here are the discussion:
* https://github.com/knative/serving/pull/9215
* https://github.com/knative/serving/issues/9077

cmd/webhook/main.go
nak3: Although it makes sense, do you know why serving's webhook is not disabling it ? - https://github.com/knative/serving/blob/main/cmd/webhook/main.go#L166-L171

ReToCode (commented):
Good points @nak3.
Other than that, LGTM

cmd/webhook/main.go
skonto: I added the default probes at the knative/pkg side and it was suggested to enable by default :shrug:. I guess I can do a follow up there too.

config/500-webhook-deployment.yaml
skonto: Yes I am aware of that was not sure to what extend it is a problem.

skonto:
@nak3 I changed the threshold values I will do a PR at the Serving side too.

config/500-controller.yaml
```diff
@@ -81,11 +81,28 @@ spec:
           seccompProfile:
             type: RuntimeDefault
 
+        readinessProbe:
+          httpGet:
+            path: /readiness
+            port: probes
+            scheme: HTTP
+          periodSeconds: 5
+          failureThreshold: 3
```
skonto, line 90: default is 3 but I want to make it explicit.

skonto:
@nak3 I have created this PR https://github.com/knative/serving/pull/13885 to address the Serving issue. We should backport downstream when all these are merged.

nak3:
Thank you. Let's wait until serving's PR was merged. We can apply the feedback into net-istio as well if necessary.

dprotaso:
FYI Serving's PR merged

config/500-webhook-deployment.yaml
```diff
@@ -82,6 +82,25 @@ spec:
...
+          failureThreshold: 6
+
```
nak3, line 103: One more thing, probably we should add `initialDelaySeconds: 20`? I think https://github.com/knative/serving/pull/9661 makes sense and other webhook have it.

nak3:
/lgtm
/approve

Thank you!

PR #1104, opened 2023-04-28
Bump k8s version on github action

nak3:
This patch bumps k8s version on CI and changes to use `chainguard-dev/actions/setup-kind@main` for setup.

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -30,18 +30,6 @@ jobs:
         istio-version:
         - latest
 
-        # Map between K8s and KinD versions.
-        # This is attempting to make it a bit clearer what's being tested.
-        # See: https://github.com/kubernetes-sigs/kind/releases
-        include:
-        - k8s-version: v1.24.7
-          kind-version: v0.17.0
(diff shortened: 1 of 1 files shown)


k4leung4 (approved):

ReToCode (commented):
Are we getting the always the latest with `.x`? Also are the image shas stable in `chainguard-dev/actions/setup-kind@main`?

It might get slightly different behaviour between builds, but I don't think it should be an issue, as these are just bugfix versions. @dprotaso, @psschwei WDYT?

Other than that, LGTM.

ReToCode:
I see, I just mentioned it because of

> Map between K8s and KinD versions.
> This is attempting to make it a bit clearer what's being tested.

But I'm fine with merging it.

dprotaso:
> So I would like to hear from Dave

It can be used in serving - i just haven't gotten around to it. ;)

I was first testing it in net-gateway-api repo as a sanity check


The images are hard coded so they're not pulling in the latest patches unless they're updated in chainguards action

https://github.com/chainguard-dev/actions/blob/main/setup-kind/action.yaml#L92


/hold cancel

dprotaso:
Also bumping the latest patch hasn't been a big issue - technically we're doing that in our e2e testing with GKE

Issue #938, opened 2022-06-07
net-istio flaky conformance tests


AngeloDanducci:
Failing tests can be found [here](https://prow.knative.dev/?job=nightly_net-istio_main_periodic)

TestIngressConformance/Update/# seems to be the primary culprit, with the issue
 `Reason:"NewObservedGenFailure", Message:"unsuccessfully observed a new generation"`
 
 Will dig into this a bit more and update the issue with any further findings.

AngeloDanducci:
Looks like it may be related to this as the observed generation isn't incrementing.
https://github.com/istio/istio/issues/31931

dprotaso:
`NewObservedGenFailure` is actually an error that the knative controllers set

https://github.com/knative/pkg/blob/d1d5c849073b4c59fabeef1aaaf5eece021d9ee2/reconciler/reconcile_common.go#L68-L72

net-istio should rebuild the status conditions - but I wonder if something is causing them to short circuit

dprotaso:
We don't use Istio's Virtual Service Status block - https://github.com/knative-sandbox/net-istio/blob/abad502be7e14ef7281de0d77dcf5525d1e7e40b/config/400-config-istio.yaml#L67-L70

Since it was really always broken lol

PR #1112, opened 2023-05-23
Disable update test

dprotaso:
This test is very flakey and hindering contributions to this repo - we need fix this but in the meantime let's disable it

See: https://github.com/knative-sandbox/net-istio/issues/938


psschwei (approved):
/lgtm
/approve

Issue #1103, opened 2023-04-28
Use ambient mesh instead of current sidecar mesh for CI


nak3:
As per title, current "mesh" test enabls sidecar on all of Knative components such as activator, autoscaller, kservice.
We should test ambient mesh instead of the sidecar mesh then make bug reports on the Istio upstream.

nak3:
The test passed - https://github.com/nak3/net-istio/pull/15
I will start this after Istio 1.18 released.

kimwnasptd:
Hey @nak3, we are evaluating in Kubeflow the support of Ambient mesh and Knative is an indirect dependency for us (for KServe).

I'm looking to play around with Knative and it's ambient mesh state. Could you help me understand the overall state with Knative and ambient mesh?

nak3:
Hi @kimwnasptd 
I have moved on to another company and project. So, could you please open a new issue and ask the current maintainers about the status?

PR #1133, opened 2023-06-16
Add ambient profile

nak3:
As per title, this patch adds ambient profile under `third_party` and github action.

Fix https://github.com/knative-sandbox/net-istio/issues/1103

### DIFF
--- a/test/config/100-test-namespace.yaml
+++ b/test/config/100-test-namespace.yaml
@@ -16,3 +16,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: serving-tests
+  labels:
+    istio.io/dataplane-mode: ambient
--- a/third_party/istio-latest/istio-kind-ambient.yaml
+++ b/third_party/istio-latest/istio-kind-ambient.yaml
@@ -1,0 +1,41 @@
+# Copyright 2020 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  profile: ambient
+  values:
+    global:
+      proxy:
+        clusterDomain: cluster.local
+    pilot:
+      autoscaleMin: 1
+      autoscaleMax: 3
+      cpu:
+        targetAverageUtilization: 60
+    gateways:
+      istio-ingressgateway:
+        autoscaleEnabled: false
+        type: NodePort
+
+  components:
+    ingressGateways:
+      - name: istio-ingressgateway
+        k8s:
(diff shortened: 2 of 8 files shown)


xtreme-vikram-yadav:
👍  Aside from the shellcheck errors, it looks good.

nak3:
Thank you! The sell check error is from the code which I haven't touched. It seems the error is surfaced due to the file change.
So I think it is better to fix it by a separate PR.

nak3:
/cc @dprotaso @psschwei @ReToCode 

Is it alright to add this ambient profile? Or do we need some discussion for this?

psschwei:
No issues adding it from my end.

third_party/library.sh
```diff
@@ -115,7 +115,16 @@ metadata:
 ---
 EOF
 
-    ${ISTIO_DIR}/bin/istioctl manifest generate -f "$file"  "$@" | add_crd_label >> "${tmpfile}"
+    if [[ "$file" == *"ambient"* ]] ; then
+      # The options are duplicated but necessary both istio.yaml and command option.
+      "${ISTIO_DIR}"/bin/istioctl manifest \
+           --set profile=ambient \
```
dprotaso, line 121: Is this necessary? You set the profile to `ambient` in the IstioOperator yaml

third_party/library.sh
```diff
@@ -115,7 +115,16 @@ metadata:
...
+	   --set components.ingressGateways[0].enabled=true \
+	   --set components.ingressGateways[0].name=istio-ingressgateway \
```
dprotaso, line 123: This can be done in the operator yaml as well

https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#IstioComponentSetSpec

third_party/istio-latest/istio-kind-ambient.yaml
```diff
@@ -0,0 +1,41 @@
...
+        clusterDomain: cluster.local
+    pilot:
+      autoscaleMin: 1
+      autoscaleMax: 3
+      cpu:
+        targetAverageUtilization: 60
+    gateways:
+      istio-ingressgateway:
+        autoscaleEnabled: false
+        type: NodePort
```
dprotaso, line 31: MetalLB works so well with kind - I don't think we want to use NodePort anymore

https://kind.sigs.k8s.io/docs/user/loadbalancer/

dprotaso:
> Is it alright to add this ambient profile?

Sure

third_party/istio-latest/istio-kind-ambient.yaml
nak3: Yes, we don't want to use NodePort. But current default one (`istio-kind-no-mesh.yaml`) also has the NodePort so if you are alright I will change it in another PR.

dprotaso: Sure

test/config/100-test-namespace.yaml
```diff
@@ -16,3 +16,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: serving-tests
+  labels:
+    istio.io/dataplane-mode: ambient
```
dprotaso, line 20: I'm assuming we need the label on the net-istio controller's namespace?

So that probing happens over the ambient mesh?

dprotaso: https://github.com/knative-sandbox/net-istio/blob/main/test/config/100-serving-namespace.yaml

nak3: Thank you. I missed it. Updated.

Issue #718, opened 2021-08-05
`config.Istio` does not need to take multiple Gateways


nak3:
The following `IngressGateways` and `LocalGateways` should be a single Gateway:

https://github.com/knative-sandbox/net-istio/blob/d7625fa51eb7e53757e4af95bc09ac86f07c921a/pkg/reconciler/ingress/config/istio.go#L89-L93

Even if it takes multiple Gateways, only one Gateway is used. For example:

https://github.com/knative-sandbox/net-istio/blob/d7625fa51eb7e53757e4af95bc09ac86f07c921a/pkg/reconciler/ingress/ingress.go#L472-L478

Also, this comment line seems strange. LocalGateway should specify the gateway url for private Ingress only.

https://github.com/knative-sandbox/net-istio/blob/d7625fa51eb7e53757e4af95bc09ac86f07c921a/pkg/reconciler/ingress/config/istio.go#L92

nak3:
Re-open this as some users misreads that multiple gateways (e.g. 2 external gw & 1 local gw like below) can be configured in net-istio.

```
data:
  gateway.ns1.gtw1: "ing1.istio-system.svc.cluster.local"
  gateway.ns2.gtw2: "ing2.istio-system.svc.cluster.local"
  local-gateway.ns3.gtw3: "ing3.istio-system.svc.cluster.local"
```

dprotaso:
> Even if it takes multiple Gateways, only one Gateway is used. For example:

😢

pastequo:
That might be naive, but why changing only the LB status method is not enough ?

cf https://github.com/knative-extensions/net-istio/pull/1218

dprotaso:
We should introduce a warning if users have configured things this way. Then maybe turn it into an error after a few releases

dprotaso:
Closing this out as we now have support for selecting  a gateway based on labels

PR #1137, opened 2023-06-27
feat: Filter on istio gateways based on ingress annotations

### LINKED ISSUE #1124: Customize Istio Virtual Service `spec.gateways` via Kservice
> Originally raised in knative/serving#14050

We have multiple Istio Ingress Gateways:
- `istio-internal` for private endpoints (accessible only via VPN)
- `istio-external` for publicly accessible endpoints

It would be great if we could customize the `spec.gateways` config in the resulting Istio Virtual Service via the Kservice manifest.
For example:
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: hello
  namespace: knative-example
spec:
  gateways:
  - istio-system/istio-external
  - istio-system/istio-internal
  template:
    spec:
      containers:
        - image: ghcr.io/knative/helloworld-go:latest
          ports:
            - containerPort: 8080
          env:
            - name: TARGET
              value: "World"
```
Could result in a Virtual Service with the following gateway config:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
[...]
  name: hello-ingress
  namespace: knative-example
spec:
  gateways:
  - istio-system/istio-external
  - istio-system/istio-internal
  hosts:
  - hello.example.com
  - hello.knative-example
  - hello.knative-example.svc
  - hello.knative-example.svc.cluster.local
[...]
```
Or:
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: hello
  namespace: knative-example
spec:
  gateways:
  - istio-system/istio-internal
  template:
    spec:
      containers:
        - image: ghcr.io/knative/helloworld-go:latest
          ports:
            - containerPort: 8080
          env:
            - name: TARGET
              value: "World"
```
Results in:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
[...]
  name: hello-ingress
  namespace: knative-example
spec:
  gateways:
  - istio-system/istio-internal
  hosts:
  - hello.example.com
  - hello.knative-example
  - hello.knative-example.svc
  - hello.knative-example.svc.cluster.local
[...]
```

/kind feature-request

pastequo:
# Changes

Add capability to restrict the list of istio gateways that will expose a knative service

- Introduce new config to attach constant, named `exposition`, to istio gateways
- The controller looks for a dedicated annotation `istio.knative.dev/exposition` on knative Ingress
- If the annotation is absent, there is no change of behavior
- If the annotation is present, only the gateways that have the annotation value (aka `exposition` value) are used
- `exposition` filters gateways that will be used for the given `visibility`

# History
- Change annotation `networking.knative.dev/istio-exposition` -> `istio.knative.dev/exposition`
- Add a check to validate config consistency. If an exposition references an unknown gateway, an error is returned
- Clarify that this notion doesn't allow an user to expand the gateway selection outside of the visibility scope
- Add clarification about a default behavior in notes

# Example

New configuration example
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-istio
  namespace: knative-serving
data:
  gateway.ns1.gtw1: "ing1.istio-system.svc.cluster.local"
  gateway.ns2.gtw2: "ing2.istio-system.svc.cluster.local"
  local-gateway.ns3.gtw3: "ing3.istio-system.svc.cluster.local"
  exposition.ns1.gtw1: "expo1"
  exposition.ns2.gtw2: "expo2,expo3"
  exposition.ns3.gtw3: "expo3"
```

Which leads to the following according to the annotation value

| Knative annotation  | visibility | gateways used | local gateways used |
| ------------- | ------------- | ------------- | ------------- |
| istio.knative.dev/exposition=expo1 | external | ns1.gtw1 | (empty) |
| istio.knative.dev/exposition=expo2 | external | ns2.gtw2 | (empty) |
| istio.knative.dev/exposition=expo3 | external | ns3.gtw3 | ns3.gtw3 |
| istio.knative.dev/exposition=expo1,expo2 | external | ns1.gtw1, ns2.gtw2 | (empty) |
| istio.knative.dev/exposition=expo-unknown | external | (empty) | (empty) |
| (empty) | external | ns1.gtw1,ns2.gtw2 | ns3.gtw3 |
| istio.knative.dev/exposition=expo1 | cluster-local | (empty*) | (empty) |
| istio.knative.dev/exposition=expo2 | cluster-local | (empty*) | (empty) |
| istio.knative.dev/exposition=expo3 | cluster-local | (empty*) | ns3.gtw3 |
| istio.knative.dev/exposition=expo1,expo2 | cluster-local | (empty*) | (empty) |
| istio.knative.dev/exposition=expo-unknown | cluster-local | (empty*) | (empty) |
| (empty) | cluster-local | (empty*) | ns3.gtw3 |

`(empty*)` in the code, for backward compatibility, the list could not be empty -> https://github.com/knative-extensions/net-istio/blob/release-1.12/pkg/reconciler/ingress/ingress.go#L180-L185

# Notes

- The code checks that the config is consistent.
If an exposition references an unknown gateway, an error is returned. In the example below, `exposition.ns-unknown.gtw-unknown` is problematic
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-istio
  namespace: knative-serving
data:
  gateway.ns1.gtw1: "ing1.istio-system.svc.cluster.local"
  local-gateway.ns2.gtw2: "ing2.istio-system.svc.cluster.local"
  exposition.ns-unknown.gtw-unknown: "expo1"
```
- The default behavior describes here https://github.com/knative-extensions/net-istio/blob/release-1.12/pkg/reconciler/ingress/ingress.go#L180-L185 also applies `visibility` filter

/kind enhancement

Fixes #1124

**Release Note**

```release-note

```

**Docs**

```docs

```

------------------------------------------------------------------------------------------------
# Initial Version

# Changes

- Introduce 2 new annotations on Ingress: `serving.knative.dev/istio-gateways` & `serving.knative.dev/istio-local-gateways`
- If an annotation is set, net-istio won't use the full list of gateway configured in the config. An **intersection** of the 2 lists will be used. 
- If TLS is specified, the custom gateway that are created specifically for wildcard also refer to istio ingress gateway filtered by the annotation

# Example

2 ingress gateway have been created while installing istio, in `istio-system` namespace
- `istio-private-ingressgateway`: This ingress is exposed on a private network
- `istio-public-ingressgateway`: This ingress is exposed on internet

The split public/private is purely informative, any gateway topology is supported

Both gateways are defined in net-istio `config-istio` ConfigMap
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-istio
  namespace: knative-serving
data:
  gateway.knative-serving.knative-private-ingress-gateway: "istio-private-ingressgateway.istio-system.svc.cluster.local"
  gateway.knative-serving.knative-public-ingress-gateway: "istio-public-ingressgateway.istio-system.svc.cluster.local"
  local-gateway....
```

While creating a knative service, the user can add an annotation to specify on which gateway the service should be exposed. Local gateways can also be filtered (not illustrated on this example)

- Only on private network:
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: my-private-service
  annotations:
    serving.knative.dev/istio-gateways: knative-serving/knative-private-ingress-gateway
```
will result for the creation of the virtual service `my-private-service-ingress`:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: my-private-service-ingress
spec:
  gateways:
  - knative-serving/knative-private-ingress-gateway
  - knative-serving/knative-local-gateway # whatever the name it has
...
```

- Only on internet:
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: my-public-service
  annotations:
    serving.knative.dev/istio-gateways: knative-serving/knative-public-ingress-gateway
```
will result for the creation of the virtual service `my-public-service-ingress`:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: my-public-service-ingress
spec:
  gateways:
  - knative-serving/knative-public-ingress-gateway
  - knative-serving/knative-local-gateway # whatever the name it has
...
```

- On both:
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: my-service
  annotations:
    serving.knative.dev/istio-gateways: knative-serving/knative-public-ingress-gateway,knative-serving/knative-private-ingress-gateway
```
or
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: my-service
```
will result for the creation of the virtual service `my-service-ingress`:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: my-service-ingress
spec:
  gateways:
  - knative-serving/knative-public-ingress-gateway
  - knative-serving/knative-private-ingress-gateway
  - knative-serving/knative-local-gateway # whatever the name it has
...
```

# Notes

- Intersection versus Directly use the list from the user annotation. In term of negative impact:
  - intersection: if the user sets an undefined gateway, the value will disappear silently
  - non-intersection: the annotation values still have to be define in the configuration to create the gateways. The user could have pre-created the gateways but net-istio is "dupplicating" the gateways for TLS. That would fail, the gateway wouldn't have any selector
- If the user doesn't set any valid value, the code would try to create a virtual service without any gateway. The virtual service is not created at all. **That could be mitigated if local-gateway can't be filtered**
- This design relies on the fact that annotation on knative service cr will be copy to knative ingress cr
- Should the code log some message if a gateway from an annotation had been ignored ? Should it raise an error instead ?

/kind enhancement

Fixes #1124

**Release Note**

```release-note

```

**Docs**

```docs

```


KauzClay:
hey @pastequo , thanks for this contribution!

> * Intersection versus Directly use the list from the user annotation. In term of negative impact:
>   * intersection: if the user sets an undefined gateway, the value will disappear silently
>   * non-intersection: the annotation values still have to be define in the configuration to create the gateways. The user could have pre-created the gateways but net-istio is "dupplicating" the gateways for TLS. That would fail, the gateway wouldn't have any selector

I'm not sure I'm following exactly, could you elaborate a bit more on the difference between intersection and non-intersection?

In both cases, it seems like it could only use whatever was defined in the configmap. You wouldn't want to have it reconciled by a gateway that knative didn't know about, right? Sorry if I'm misinterpreting things.

> * If the user doesn't set any valid value, the code would try to create a virtual service without any gateway. The virtual service is not created at all. That could be mitigated if local-gateway can't be filtered

I may be misunderstanding, but my instinct would be to fall back to the current default behavior of applying every gateway that is listed in the configmap. That way it wouldn't really change anything for people not using this annotation/use case.

> Should the code log some message if a gateway from an annotation had been ignored ? Should it raise an error instead ?

Maybe we could update the status on the Kingress with a condition explaining this? Though idk if this is too ingress-specific to make its way up into Knative core stuff?

--- 

Also, one of things I was thinking about with this feature is that, if you use different gateways, I think you also need to consider what domains are being used. I think in @rblaine95's case, they wanted to use the same domain for both gateways, so it is less of a problem. 

But could there be a scenario where DNS is set up to map one domain to gateway A and another to gateway B? Then you'd need to make sure the Ksvc with the gateway A annotation has the right labels so that the domain it is given is correct.

In this case, I'm thinking this might be a hard error to discover. Idk if there would be a way to make net-istio warn you that the domain on a ksvc isn't served by the gateway it is tied to, since that info kinda lies outside of Knative.

pastequo:
Thanks for your feedbacks !

# Check user annotation

non-intersection vs intersection was not very clear, let me rephrase it:
Should the code check that the gateways set in annotations are defined in `config-istio` configmap ? If yes, what to do in case of an invalid value ? (see next section)

The TLS part of the code required gateways to be defined in the `config-istio` configmap, the non-TLS part doesn't

# Invalid value management

Let's enumerate the different possibilities and merge into it the "logging question".

A value in the annotations can be invalid for 2 reasons:
- it's not well formatted (the code is expected a qualified name in the form of `$namespace/$gateway-name`)
- `config-istio` configmap doesn't reference this gateway

I'm assuming the code should respond the same way if a value is invalid, whatever the reason.

From my understanding, when reconciling an ingress, the code is merging the "gateways" with the "local-gateways" and put the result in the virtual service named `$serviceName-ingress`. For the TLS specific path, only the "gateways" are considered.

## All values are valid (well-formatted & present in `config-istio` configmap)

Happy path, just apply the config

## There is no annotation

Still happy path, default to the list in the `config-istio` configmap

## All values are invalid (either not well-formatted or not present in `config-istio`)

It either means:
- both "local" and "non-local" annotations only contain invalid values
- "non-local" annotation only contain invalid values & TLS is activated

What is the expected behavior:
- Should the reconciliation fail ?
- Should it default to the list in the `config-istio` configmap ? _(As a user, I would find this behavior very surprising)_
- Should the virtual services be still created with invalid values ?
- How to surface the error(s) ?

## Some values are invalid, some are valid

- Should the reconciliation fail ?
- Should the configuration be partly applied (with only the right value) ? 
- Should it default to the list in the `config-istio` configmap ? _(As a user, I would find this behavior very surprising)_
- - Should the virtual services be still created with invalid values ?
- How to surface the error(s) ?

# Local gateways

I definitely lack knowledge around the concept of "local-gateways", I don't see what is their purpose. So I don't know if they should be "select-able" as well with annotations. Strictly speaking about the original issue(s), I don't think it's required

# DNS consideration

Regarding the DNS part, my first instinct would be to say it's out of scope.

In the `config-domain` configmap, something like that could be defined
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-domain
  namespace: knative-serving
data:
  my-public-domain.com: |
    selector:
      traffic: public
  my-private-domain.com: |
    selector:
     traffic: private
```
Then I think it's up to the end users to ensure they are creating something consistent, not like below
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: my-badly-configured-service
  annotations:
    serving.knative.dev/istio-gateways: knative-serving/knative-public-ingress-gateway # public here
  labels:
    traffic: private # private here
spec:
```

But it could be a nice information to surface

pastequo:
I would argue that as-of-today the default behavior is to expose the service to all gateways defined in the configmap. Therefore this PR can be viewed as a way to define where you don't want to expose the service, meaning it filtered what the admin defined, not overwrite.

I understand your point about  

> I don't know whether letting users run services that are 'unexpectedly' accessible in intranet networks is a risk

But letting users run services that are 'unexpectedly' accessible from internet (and more generally from any unexpected networks) is a bigger risk imho. I think the right approach here is to not split by "public/private" but generalize to "different networks" and let the admin, thru the ingress gateway definition, put any meaning on them. Then I think it's up to the service owner to know & specify from which network their service should be accessible

Regarding the namespaces approach, I, as a service owner, am personally used to have the flexibility to define how to split my namespace. which is convenient for RBAC, networking policies, etc. If I was asked to use a _static_ list of namespace, that would probably imply more problem.

Your proposal didn't state it has to be _static_ but if the list is _dynamic_ (for example deploy on namespace `network1-*` to expose to `network1`) than the proposal is equivalent to the label/annotation one, with an additional layer of complexity. Instead of choosing a annotation/label, the user is choosing a namespace

Just my 2 cents.

dprotaso:
I think the gateway api sets a good precendent here (long term we'll want to consolidate around it) 

They make this an operator concern and the gateway resource defines which routes can attach to itself

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.AllowedRoutes

Thus the operator specifies which namespaces are allowed to use certain gateways and then the user can specify the gateways that it wants

pastequo:
Ok, if I understand you correctly, there are 2 different functionalities to consider:
- User should be able to specify which gateway they want to use (as long as they are defined in the config)
- Operator should be able to restrict gateway access (probably to certain namespace)

The current status is that all gateways are always used. This PR allows to achieve the 1st functionality without, I think, putting constraint on the 2nd functionality.

Is there a reason to achieve the 2 functionalities simultaneously? And even in the same PR?

dprotaso:
Following up from our discussion in the Serving WG group.

We identified this problem sorta fits into the `visibility` feature we have for our services. This is where a user can specify using an annotation a visibility for their Knative Service - eg. `networking.knative.dev/visibility=cluster-local`

Currently we associate `cluster-local` visibility to gateways with the `local-gateway.` prefix in the config-istio configmap. Similarly, external Gateways with the `gateway.` prefix.

> User should be able to specify which gateway they want to use (as long as they are defined in the config)

I think we shouldn't expose more to the user than what's necessary - it's the wrong level of abstraction for the user.

Hence I think a mechanic for the operator to define a new visibility in `config-istio` and then users can use this higher level constant.

A rough idea could be

```yaml
gateway.knative-serving.knative-ingress-gateway: "istio-ingressgateway.istio-system.svc.cluster.local"
gateway-visibility.vpc.knative-serving.knative-ingress-gateway: "istio-ingressgateway.istio-system.svc.cluster.local"
gateway-visibility.vpc.knative-serving.knative-vpc-gateway: "istio-ingressgateway.istio-system.svc.cluster.local"
```
Thus if the user specifies the annotation`networking.knative.dev/visibility=vpc` - it will look up and attach the virtual service to Istio Gateways `knative-ingress-gateway` & `knative-vpc-gateway`

Thus the template is
```yaml
gateway-visibility.{visibility}.{gateway-ns}.{gateway-name}: "{target-service-for-probing}"
```

> Operator should be able to restrict gateway access (probably to certain namespace)

I think you're right about me introducing an additional constraint. We can always add this later.

cc @nak3 @ReToCode @skonto for thoughts/ideas on alternative approaches?

ReToCode:
> cc @nak3 @ReToCode @skonto for thoughts/ideas on alternative approaches?

Just some thoughts so far:
I'm still struggling a bit with the mix of which gateway a Knative Service should be attached to and the current behaviour of `visibility=cluster-local` which is used to differentiate between domains. It feels like we are mixing two things to me.

Also currently, only istio can even do the multi-gateway stuff, so it is a bit strange to add this as public interface to a Knative Service. So I feel this should be configured by an admin and only handled by `net-istio` in `config-istio` (at least for now, gateway-api will probably change this).

Currently we have two gateway types and attach istio resources to all gateways with the matching type, but not all domains:
- cluster-local = cluster local doman `<svc>.<ns>.svc.cluster.local`
- cluster-external = domain configured in `config-domain` (+ domain mappings)

So what would be the domain for `vpc` or any other visibility? So I feel we should keep those two types of gateways and domains and try to only target the gateway selection.

What about a new config that has two label selectors (roughly):
```yaml
- serviceSelector:
   matchLabels:
     my-workload-selector: private
  gatewaySelector:
    matchLabels:
      gateway-visibility: vpc
- serviceSelector:
   matchLabels:
     my-workload-selector: public
  gatewaySelector:
    matchLabels:
      gateway-visibility: external
```
Then people can add labels on gateways + services as they need. `net-istio` would still have two gateway types, but filter out the gateways it attaches the istio resources to.

nak3:
I think both Dave's and Reto's idea sounds good. 

With said, Reto's will introduce a brand new config against the current config (=`gateway.knative-serving.knative-ingress-gateway: "istio-ingressgateway.istio-system.svc.cluster.local"`) so the migration will be complicated, I think.

Just FYI, we had a proposal about expanding visibility https://github.com/knative/serving/issues/6642 before though it cannot be a short term solution.

ReToCode:
> With said, Reto's will introduce a brand new config against the current config (=gateway.knative-serving.knative-ingress-gateway: "istio-ingressgateway.istio-system.svc.cluster.local") so the migration will be complicated, I think.

How do you mean that? The current config of gateways would stay the same in my proposal. I would just add a new and optional key to influence the gateway <> service selection.

nak3:
I see. Then, user's migration is not necessary.

But still, the code change (`parseGateways()` in `pkg/reconciler/ingress/config/istio.go`) may become a little bit complicated compared to the prefix based parse (`gateway-visibility.{visibility}.{gateway-ns}.{gateway-name}:`).

pastequo:
I'm back after some PTO and still willing to move forward with this PR !

To sum up the different messages from @dprotaso @ReToCode & @nak3 
- Users shouldn't reference directly istio gateways
- This PR should not be mixed with the `visibility` feature
- It should avoid a mandatory user migration
- Config should be easy to parse with the current code

IMHO, I think
- it's more appropriate for the user to use an annotation rather than a label 
- Istio gateways shouldn't be referenced by different ways in the config (the current config already use qualified name, I wouldn't add labels to reference the same thing)

So, if we agree on that, by mixing the different ideas, another proposal. Basically it's the same as my first proposal without exposing the gateway names:
- `net-istio` config introduces a new config to associate istio gateway with a constant (that I named "exposition")
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-istio
  namespace: knative-serving
data:
  gateway.knative-serving.knative-internal-ingress-gateway: "istio-internal-ingressgateway.istio-system.svc.cluster.local"
  gateway.knative-serving.knative-external-ingress-gateway: "istio-external-ingressgateway.istio-system.svc.cluster.local"
  local-gateway.knative-serving.knative-local-gateway: "istio-internal-ingressgateway.istio-system.svc.cluster.local"
  exposition.knative-serving.knative-internal-ingress-gateway: "my-value-1"
  exposition.knative-serving.knative-external-ingress-gateway: "my-value-2,my-value-3"
```
Thus the template is 
`exposition.{gateway-namespace}.{gateway-name}: "{list-of-value-separated-by-a-comma}"`

- Users specifies the annotation `networking.knative.dev/istio-exposition=my-value-1`
- The code adds the new field in the struct
```go
// Gateway specifies the name of the Gateway and the K8s Service backing it.
type Gateway struct {
	Namespace  string
	Name       string
	ServiceURL string
	Exposition sets.String
}
```

Any thoughts ?

I'm available to work on whatever implementation is chosen

nak3:
@pastequo Thank you so much! Could you update the initial description https://github.com/knative-extensions/net-istio/pull/1137#issue-1776812747 if it was changed? I guess that the latest commit is different from the initial proposal e.g. annotations, correct?

nak3:
> Should it raise an error while parsing the configuration ?

"parsing the configuration" means `parseExposition()`? If so, I think yes.

> Should the filtering also apply to the local gateways ?

I think either is fine for now but we should implement this with the assumption to add the local gateway later.

I guess we need to support it when users use multi-tenant ServiceMesh in the future - https://istio.io/latest/docs/ops/deployment/deployment-models/#multiple-clusters

> Should it raise a reconciliation error if the exposition specify in the ingress annotation is unknown (not present in the config)

No, I think we should not raise a reconcilation error but we should conisder to add a validation error via webhook. (I am still thinking about it though.)

dprotaso:
/hold 

release is tomorrow - will unhold afterwards

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -45,6 +45,10 @@ const (
 	GatewayHTTPPort       = 80
 	dns1123LabelMaxLength = 63 // Public for testing only.
 	dns1123LabelFmt       = "[a-zA-Z0-9](?:[-a-zA-Z0-9]*[a-zA-Z0-9])?"
+
+	// Ingress annotation to select gateways thru exposition flag
+	// public for testing purposes
+	ExpositionAnnotation = "serving.knative.dev/istio-exposition"
```
nak3, line 51:
```suggestion
	ExpositionAnnotation = "networking.knative.dev/istio-exposition"
```

nak3:
Hi @pastequo I just tried this to understand the implementation but the status becomes `meshOnly: true`. I think my configuration somehow is wrong but is it possible to add an e2e test under `test/e2e/`? Then, I think I understand what was wrong.

```
$ kubectl get king  -o yaml hello-example
  ...
  status:
    conditions:
   ...
    privateLoadBalancer:
      ingress:
      - meshOnly: true
    publicLoadBalancer:
      ingress:
      - meshOnly: true
```

pkg/reconciler/ingress/resources/gateway.go
nak3: Or `istio.knative.dev/istio-exposition`. Kourier uses `kourier.knative.dev/`

https://github.com/knative-extensions/net-kourier/blob/main/pkg/config/config.go#L66

pastequo:
> Hi @pastequo I just tried this to understand the implementation but the status becomes `meshOnly: true`. I think my configuration somehow is wrong but is it possible to add an e2e test under `test/e2e/`? Then, I think I understand what was wrong.
> 
> ```
> $ kubectl get king  -o yaml hello-example
>   ...
>   status:
>     conditions:
>    ...
>     privateLoadBalancer:
>       ingress:
>       - meshOnly: true
>     publicLoadBalancer:
>       ingress:
>       - meshOnly: true
> ```

I don't reproduce locally what you observed

- I changed the annotation name
- I made `NewIstioFromConfigMap` return an error if the configuration is not consistent
- I updated the PR description accordingly
- I added a e2e test. Let me know if it's done the right way. We will see how the pipeline behaves...

(On a side note, it wasn't easy to setup a local environment to replay e2e... It might be worth documenting how you proceed. I saw some weird things too that might require clean up. For example github workflows references `CONFIG_ISTIO="./third_party/istio-${{ matrix.istio-version }}/istio-kind-${{ matrix.istio-profile }}/config-istio.yaml"` but the files are not named `config-istio.yaml` but `istio.yaml`. If I'm correct there isn't any `config-istio.yaml` )

nak3:
Thank you. I understand that the `MeshOnly: true` happens when no LB matched. Is it possible to set `MarkLoadBalancerNotReady` instead of `MeshOnly: true`  when there is no matched LB?

pastequo:
Wouldn't it be more confusing ? By reading `MarkLoadBalancerNotReady` I expect to have a LB that will eventually become ready

nak3:
> Wouldn't it be more confusing ? By reading MarkLoadBalancerNotReady I expect to have a LB that will eventually become ready

Yes, and users will take a look at their LB (=Istio Gateway) and realize that the LB does not exist due to their wrong configuration, won't they?

By the way, I realized that the designe has some problems:

(copied from https://github.com/knative-extensions/net-istio/pull/1137#issue-1776812747)

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-istio
  namespace: knative-serving
data:
  gateway.ns1.gtw1: "ing1.istio-system.svc.cluster.local"
  gateway.ns2.gtw2: "ing2.istio-system.svc.cluster.local"
  ...
```

You assume that `ns1.gtw1` & `ns2.gtw2` are used when annotation is `empty`. But serving's prober does not check if both are available. So, multiple gateways for one ksvc should have a problem.
So I think we still need a template `gateway-visibility.{visibility}.{gateway-ns}.{gateway-name}: "{target-service-for-probing}"` to support.

pastequo:
- Regarding the MarkLoadBalancerNotReady status, I think the first question to answer is to agree if it is valid or not if there is no gateway (local or not) for a given visibility. If not then we need a mechanism to surface the error. If it is valid, then I think there is no need to mark the status as MarkLoadBalancerNotReady since all the expected LBs are ready (in this case 0, only the mesh is configured)

BTW, there is a subtlety between the istio ingress gateways and the istio gateways. The LBs (exposing Istio *ingress* gateway, aka envoy) are ready and healthy. But none of them are used because there is no istio gateways (a gathering of configuration) for this ksvc. So it could be confusing for the user, because he shouldn't look at the LB nor the istio ingress gateways

- Regarding the probes, after a quick look at the code, I think all the endpoints are checked, but only the first one is reported. I think this issue is independent from my PR. As of today - without this PR - I can configure multiple gateways in my configuration. The capability to filter on those gateways shouldn't influence the probes. It might be better to handle the 2 topics separately

ReToCode:
I still think it is not the best idea to mix the current visibility with the new gateway selection (https://github.com/knative-extensions/net-istio/pull/1137#issuecomment-1655211889). Similarly, [this](https://github.com/knative-extensions/net-istio/pull/1137#issuecomment-1794708869) goes the same direction.

pastequo:
Thanks for the reply @ReToCode.

I can confirm you that I took into account your remarks: the 2 notions (visibility & exposition) are **not** mixed.
What I designed and I hopefully implemented, is to limit the list of gateway per visibility without changing the logic on top of that that choses which list of gateway to use (based on visibility)

I will edit the design to state it clearly

dprotaso:
Funny enough @nak3's issue (https://github.com/knative-extensions/net-istio/issues/718) made me realize there's a lot more work to support this feature properly - so that's definitely something we'd want to address.

Secondly, I wonder if it would make sense to take inspiration from config-domain changes that are likely to land - https://github.com/knative/serving/pull/14543 Since it seems there's still a bit of contention on how to structure config-istio.

What I like about that Serving PR is that it supports a new config format and the old one. Given that it might be best rethink what we want config-istio to look like instead of trying to shoe horn it into the existing format.

eg. we could have
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-istio
  namespace: knative-serving
data:
  internal-gateways: 
    - service-namespace: istio-system
      service-name: knative-cluster-local
  external-gateways: 
    - service-namespace: istio-system
      service-name:  ing2
    - service-namespace: istio-system
      service-name:  ing1      
```

and then introduce the concept of selectors for services and even namespace selectors

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-istio
  namespace: knative-serving
data:
  internal-gateways: |
    - namespace: istio-system
      name: knative-cluster-local
  external-gateways: |
    - namespace: istio-system
      name:  ing2
      namespaceSelector: |
        kubernetes.io/metadata.name: blah
    - service-namespace: istio-system
      service-name:  ing1      
      selector: |
         app: foo      
```

We could even extend this to support matching gateways depending on domain so you don't have to repeat the namespace selector in two configs

Thoughts on this approach @ReToCode & @nak3 ?

@pastequo would the above give you the flexibility you require?

It might make sense to move this proposal to a shared google document rather than a PR - @pastequo can you open an issue regarding this incase you haven't done so?

ReToCode:
@dprotaso I like that approach, it is closer to what we actually want to configure and does not mix with the visibility (as per https://github.com/knative-extensions/net-istio/pull/1137#issuecomment-1655211889).

So you'd add this as an additional config to keep being reverse compatible with the new structure to have precedence?

Side-note: probably we also want to rename the `internal-gateways` to `local-gateways`.

pastequo:
> Funny enough @nak3's issue (#718) made me realize there's a lot more work to support this feature properly - so that's definitely something we'd want to address.

I don't have the history you have on this project, but it seemed to me that there wasn't a lot of work for that. All the code is plugged to work with a list of gateway except for the LB Status methods that already accept an array.

> Secondly, I wonder if it would make sense to take inspiration from config-domain changes that are likely to land - https://github.com/knative/serving/pull/14543 Since it seems there's still a bit of contention on how to structure config-istio.

Should it block this PR ? Reworking the config structure seems independent to me. Tho I understand that might be a good opportunity.

> @pastequo would the above give you the flexibility you require?

It seems so but some remarks:
- Even if the "admin" should have capabilities to restrict which services can be exposed, I believe the "user" should be the one who in the first place explicit their requirement about how to expose their service. To achieve that, I think annotation would be more appropriate than labels, as the user wants to trigger a special behavior. So the configuration should expose a way to associate a special annotation value to a internal/external-gateway.
-  `selector` sounds as a "labels-based" approach to me, it can be kept if we want to offer to the "admin" the capability to filter by labels
- Similarly to what I did in my description, the important part is to define what is the expected behavior when no labels/annotation matches, when the user specifies nothing, etc

> It might make sense to move this proposal to a shared google document rather than a PR - @pastequo can you open an issue regarding this incase you haven't done so?

I was referencing another user issue: https://github.com/knative-extensions/net-istio/issues/1124
Would it be ok for you or should I create something else specific ?

pastequo:
Google document added

https://docs.google.com/document/d/12X1-9nhjAhpf-Wlt3RbdoMbvx31zFBva/edit#heading=h.gjdgxs

PR #1218, opened 2023-11-27
fix: Return LB status for each gateway

### LINKED ISSUE #718: `config.Istio` does not need to take multiple Gateways

pastequo:
# Changes

- :bug: Set LB status for all gateways

/kind bug

Fixes #718

Note that the issue rather recommend to avoid setting multiple gateway in the config.

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -232,8 +232,16 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	}
 
 	if ready {
-		publicLbs := getLBStatus(publicGatewayServiceURLFromContext(ctx))
-		privateLbs := getLBStatus(privateGatewayServiceURLFromContext(ctx))
+		publicLbs, err := getLBStatuses(ctx, v1alpha1.IngressVisibilityExternalIP)
+		if err != nil {
+			return fmt.Errorf("failed to get external LB status: %w", err)
+		}
+
+		privateLbs, err := getLBStatuses(ctx, v1alpha1.IngressVisibilityClusterLocal)
+		if err != nil {
+			return fmt.Errorf("failed to get local LB status: %w", err)
+		}
+
 		ing.Status.MarkLoadBalancerReady(publicLbs, privateLbs)
 	} else {
 		ing.Status.MarkLoadBalancerNotReady()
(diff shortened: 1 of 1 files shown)


pkg/reconciler/ingress/ingress.go
```diff
@@ -232,8 +232,16 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
...
 
 	if ready {
-		publicLbs := getLBStatus(publicGatewayServiceURLFromContext(ctx))
-		privateLbs := getLBStatus(privateGatewayServiceURLFromContext(ctx))
+		publicLbs, err := getLBStatuses(ctx, v1alpha1.IngressVisibilityExternalIP)
+		if err != nil {
+			return fmt.Errorf("failed to get external LB status: %w", err)
+		}
+
+		privateLbs, err := getLBStatuses(ctx, v1alpha1.IngressVisibilityClusterLocal)
```
nak3, line 240: In my understanding, this will set multiple gateways to the LB status when users configured multiple gateways. Did that work with serving? For example, here is the code which disallows it - 
https://github.com/knative/serving/blob/9b0d1851055e91ced3965dc3cb8d9edb08f833ee/pkg/reconciler/route/resources/service.go#L87-L91

pastequo: Ok, I wasn't aware of such limitation at serving side.

Small remark tho, we can have multiple istio gateways that point to the same istio ingress gateway (and therefore LB)

PR #1221, opened 2023-11-30
Update net-istio to use `KnativeCertificate` instead of control-protocol secret

ReToCode:
# Changes
- https://github.com/knative-extensions/net-istio/pull/1085 introduced support for `system-internal-tls` in net-istio, this PR makes it work with the changes in https://github.com/knative/serving/pull/14610
- Instead of using the secret from the custom reconciler, its now using a `KnativeCertificate`

Partially https://github.com/knative-extensions/net-istio/issues/1063

/assign @skonto 
/assign @nak3


dprotaso:
/retest

> TestIngressConformance/3/httpoption

PR #1228, opened 2023-12-19
Host cluster-local-domain-tls on cluster-local gateway with SNI

ReToCode:
# Changes
- Add cluster-local https servers when `cluster-local-domain-tls` is enabled and a KSVC exists
- Uses SNI to host multiple certificates for all cluster-local domain Knative Services

/kind enhancement

Fixes https://github.com/knative/serving/issues/14374

**Release Note**

```release-note
net-istio now hosts an additional gateway with TLS for each Knative Service when `cluster-local-domain-tls` is enabled. Note: this is an experimental alpha-feature.
```

**Docs**

Will be done once the features is complete

### DIFF
--- a/config/203-local-gateway.yaml
+++ b/config/203-local-gateway.yaml
@@ -57,4 +57,4 @@ spec:
       targetPort: 8081
     - name: https
       port: 443
-      targetPort: 8443
+      targetPort: 8444
--- a/config/203-local-gateway.yaml
+++ b/config/203-local-gateway.yaml
@@ -55,3 +55,6 @@ spec:
     - name: http2
       port: 80
       targetPort: 8081
+    - name: https
+      port: 443
+      targetPort: 8443
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -57,10 +57,9 @@ import (
 )
 
 const (
-	virtualServiceConditionReconciled = "Reconciled"
-	virtualServiceNotReconciled       = "ReconcileVirtualServiceFailed"
-	notReconciledReason               = "ReconcileIngressFailed"
-	notReconciledMessage              = "Ingress reconciliation failed"
+	virtualServiceNotReconciled = "ReconcileVirtualServiceFailed"
+	notReconciledReason         = "ReconcileIngressFailed"
+	notReconciledMessage        = "Ingress reconciliation failed"
 )
 
 // Reconciler implements the control loop for the Ingress resources.
@@ -106,8 +105,8 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	logger := logging.FromContext(ctx)
 
 	// We may be reading a version of the object that was stored at an older version
-	// and may not have had all of the assumed defaults specified.  This won't result
-	// in this getting written back to the API Server, but lets downstream logic make
+	// and may not have had all the assumed defaults specified.  This won't result
+	// in this getting written back to the API Server, but let's downstream logic make
 	// assumptions about defaulting.
 	ing.SetDefaults(ctx)
 
@@ -118,9 +117,9 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	gatewayNames[v1alpha1.IngressVisibilityClusterLocal] = qualifiedGatewayNamesFromContext(ctx)[v1alpha1.IngressVisibilityClusterLocal]
 	gatewayNames[v1alpha1.IngressVisibilityExternalIP] = sets.New[string]()
 
-	ingressGateways := []*v1beta1.Gateway{}
-	if shouldReconcileTLS(ing) {
-		originSecrets, err := resources.GetSecrets(ing, r.secretLister)
+	externalIngressGateways := []*v1beta1.Gateway{}
+	if shouldReconcileExternalDomainTLS(ing) {
+		originSecrets, err := resources.GetSecrets(ing, v1alpha1.IngressVisibilityExternalIP, r.secretLister)
 		if err != nil {
 			return err
 		}
@@ -144,7 +143,8 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 		}
 
 		nonWildcardIngressTLS := resources.GetNonWildcardIngressTLS(ing.GetIngressTLSForVisibility(v1alpha1.IngressVisibilityExternalIP), nonWildcardSecrets)
-		ingressGateways, err = resources.MakeIngressTLSGateways(ctx, ing, nonWildcardIngressTLS, nonWildcardSecrets, r.svcLister)
+		externalIngressGateways, err = resources.MakeIngressTLSGateways(ctx, ing, v1alpha1.IngressVisibilityExternalIP,
+			nonWildcardIngressTLS, nonWildcardSecrets, r.svcLister)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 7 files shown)


ReToCode:
Did some testing with Serving @ https://github.com/knative/serving/pull/14610, looks good. Test setup [here](https://github.com/ReToCode/knative-encryption/blob/main/7-cert-manager-only/README_ISTIO.md).

@skonto PTAL. Note: I'm fine with waiting until Serving is merged, then we can add this here and also do the full test-set with with net-istio in Serving.

ReToCode:
/unhold

Serving PR has been merged, @skonto PTAL.

config/203-local-gateway.yaml
```diff
@@ -55,3 +55,6 @@ spec:
     - name: http2
       port: 80
       targetPort: 8081
+    - name: https
+      port: 443
+      targetPort: 8443
```
dprotaso, line 60: To confirm my understanding - the two istio gateways we have actually use the same envoy instance but we use different ports for tenancy

For external ingress we open 
- port 80
- port 443 dynamically when we have TLS certs 

For internal traffic we have
- port 8081 (this service exposes it on port 80)
- port 8443 (this service exposes it on port 443)

dprotaso: I see 443 server being added here for external traffic:

https://github.com/knative-extensions/net-istio/blob/6dd62fe525872f7b714bb742a8a7ecab3c9b5319/pkg/reconciler/ingress/resources/gateway.go#L321-L323

dprotaso: But I don't see us adding TLS servers on port 8443 for cluster local traffic

ReToCode: Unfortunately istio is quite confusing in that config:

- Istio has an already open and exposed port 8443 for https (https://github.com/knative-extensions/net-istio/blob/main/third_party/istio-latest/istio-ci-no-mesh/istio.yaml#L9955)
- traffic from external (LB) will end up on 8443 (https://github.com/knative-extensions/net-istio/blob/main/third_party/istio-latest/istio-ci-no-mesh/istio.yaml#L10240)
- The change here will just forward traffic on the local gateway service to the same port
- Istio will use SNI to determine which route to take, if we look at the envoy config for two Knative Services we get

```
istioctl proxy-config listeners -n istio-system deploy/istio-ingressgateway
ADDRESSES PORT  MATCH                                                                            DESTINATION
0.0.0.0   8080  ALL                                                                              Route: http.8080
0.0.0.0   8081  ALL                                                                              Route: http.8081
0.0.0.0   8443  SNI: helloworld.second,helloworld.second.svc,helloworld.second.svc.cluster.local Route: https.443.second/helloworld:0.helloworld-975966116.second
0.0.0.0   8443  SNI: helloworld.second.172.17.0.100.sslip.io                                     Route: https.443.second/helloworld:0.helloworld-3797421420.second
0.0.0.0   8443  SNI: helloworld.first,helloworld.first.svc,helloworld.first.svc.cluster.local    Route: https.443.first/helloworld:0.helloworld-975966116.first
0.0.0.0   8443  SNI: helloworld.first.172.17.0.100.sslip.io                                      Route: https.443.first/helloworld:0.helloworld-3797421420.first
0.0.0.0   15021 ALL                                                                              Inline Route: /healthz/ready*
0.0.0.0   15090 ALL                                                                              Inline Route: /stats/prometheus*
```

So basically, everything https will be on 8443 and routed to 443 to the Knative Service.

- I'm not even sure if we would need to specify the port or not, but that 443 just maps to the containers 8443 port.

ReToCode: I assume we could use a different port for the internal TLS traffic, but not sure what we would gain. As you said, it's a shared component without tenancy anyway.

dprotaso: It's using tenancy via port isolation so external traffic cannot hit the 'local' gateway ports.

Essentially these ports are public - so we should use a different port for internal tls - otherwise we risk exposing private Knative Services to the internet.

https://github.com/knative-extensions/net-istio/blob/54d8f412d02a18a720c6be67cdc1372f4d4b7783/third_party/istio-latest/istio-ci-no-mesh/istio.yaml#L10232-L10247

dprotaso: Maybe this can be configurable after this lands -  https://github.com/knative-extensions/net-istio/pull/1249

ReToCode: > Essentially these ports are public - so we should use a different port for internal tls - otherwise we risk exposing private Knative Services to the internet.

That's a valid point. I'll take a look tomorrow.

ReToCode: I updated the PR to use a dedicated port for cluster-local-tls (8444). With that I get the following envoy config:

```bash
istioctl proxy-config listeners -n istio-system deploy/istio-ingressgateway
ADDRESSES PORT  MATCH                                                                               DESTINATION
0.0.0.0   8080  ALL                                                                                 Route: http.8080
0.0.0.0   8081  ALL                                                                                 Route: http.8081
0.0.0.0   8443  SNI: helloworld.default.172.17.0.100.sslip.io                                       Route: https.443.default/helloworld:0.helloworld-3797421420.default
0.0.0.0   8444  SNI: helloworld.default,helloworld.default.svc,helloworld.default.svc.cluster.local Route: https.8444.default/helloworld:0.helloworld-975966116.default
0.0.0.0   15021 ALL                                                                                 Inline Route: /healthz/ready*
0.0.0.0   15090 ALL                                                                                 Inline Route: /stats/prometheus*
```

As 8444 is not exposed, it is now only available cluster locally.

skonto:
As discussed with Reto the landing page of this repo is not ideal. There is no doc explaining what is supported and how. For example here we are moving to a design choice of having one gateway per svc for cluster local stuff but it is not captured somewhere etc. I think we need to spend some time adding content here.

pkg/reconciler/ingress/ingress.go
```diff
@@ -410,16 +436,16 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ing *v1alpha1.Ingress) pk
 		}
 	}
 
-	return r.reconcileDeletion(ctx, ing)
+	return r.cleanupCertificateSecrets(ctx, ing)
 }
 
-func (r *Reconciler) reconcileDeletion(ctx context.Context, ing *v1alpha1.Ingress) error {
-	if !shouldReconcileTLS(ing) {
+func (r *Reconciler) cleanupCertificateSecrets(ctx context.Context, ing *v1alpha1.Ingress) error {
```
skonto, line 442: Maybe `deleteAllDomainCertificateSecrets`.

pkg/reconciler/ingress/ingress.go
```diff
@@ -118,9 +117,9 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	gatewayNames[v1alpha1.IngressVisibilityClusterLocal] = qualifiedGatewayNamesFromContext(ctx)[v1alpha1.IngressVisibilityClusterLocal]
 	gatewayNames[v1alpha1.IngressVisibilityExternalIP] = sets.New[string]()
 
-	ingressGateways := []*v1beta1.Gateway{}
-	if shouldReconcileTLS(ing) {
-		originSecrets, err := resources.GetSecrets(ing, r.secretLister)
+	externalIngressGateways := []*v1beta1.Gateway{}
```
skonto, line 120: `var externalIngressGateways []*v1beta1.Gateway` or we want an empty slice instead?

skonto:
minor comments, LGTM. We can stamp if @dprotaso has no other comments.

PR #1250, opened 2024-01-26
Feat/gateway filtering

### LINKED ISSUE #1248: Filtering on external istio gateway
This issue comes from this design document: https://docs.google.com/document/d/12X1-9nhjAhpf-Wlt3RbdoMbvx31zFBva/edit#heading=h.gjdgxs

Developpers should be able to put labels on their knative service so that their service will be exposed by a specific istio ingress gateway. Operators defines the mapping between the labels and the istio ingress gateway

This issue depends on the new configuration format https://github.com/knative-extensions/net-istio/issues/1247

See the design document for more details

pastequo:
# Changes

This PR was built on top of https://github.com/knative-extensions/net-istio/pull/1249

This is why there are a lot of changes (it includes 2 PRs in one)

- Introduce LabelSelector on istio config
- Filter gateway to use based on ingress labels & configuration label selector

/kind enhancement

Fixes #1248 

**Release Note**

```release-note
Services can be exposed by a specific istio gateway. Administrators, by using the new istio config format, are able to associate a label selector to their istio gateway. More information [here](https://docs.google.com/document/d/12X1-9nhjAhpf-Wlt3RbdoMbvx31zFBva/edit#heading=h.gjdgxs)
```

**Docs**

```docs

```


pastequo:
Contrary to what was documented, I felt like it was wrong to implement here the check about the number of gateway (for example a maximum of one external gateway should be selected thru labels)

Even if it's not officially supported, a great part of the code acts as multiple gateway is possible. In particular, a lot of tests are using a configuration with multiple gateways.

If the change is confirmed, I think it should be done on a separated PR

dprotaso:
You'll need to rebase since the prior PR merged.

> I felt like it was wrong to implement here the check about the number of gateway (for example a maximum of one external gateway should be selected thru labels)

I'll bring it up at the WG group meeting tomorrow to see what others think.

My thoughts are given that we don't probe _all_ gateways you could say allowing that behaviour is broken. If we have a plan to address the broken behaviour then we can drop the check.

dprotaso:
Following up from the WG discussion - we'll want to add a check to ensure only a single gateway is selected (cause that's all we support). It's fine that this is done in a separate PR

dprotaso (commented):
~Something...~ posted the question inline in the review

pkg/reconciler/ingress/config/istio.go
```diff
@@ -207,10 +231,6 @@ func parseNewFormat(configMap *corev1.ConfigMap) (*Istio, error) {
 		ret.LocalGateways = localGateways
 	}
 
-	if len(ret.LocalGateways) > 1 {
```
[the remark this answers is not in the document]
dprotaso: Unless you're supporting multiple gateways - which is a question I asked here: https://github.com/knative-extensions/net-istio/pull/1250#discussion_r1496561458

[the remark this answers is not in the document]
dprotaso: Whatever the decision we should make sure we update the docs in https://github.com/knative-extensions/net-istio/blob/133392a7ea3295b281c51f49a4a943819f689c08/config/400-config-istio.yaml#L107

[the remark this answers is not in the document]
pastequo: I added the check, so I think the doc is ok

pkg/reconciler/ingress/config/istio_test.go
```diff
@@ -438,6 +416,106 @@ func TestGatewayConfiguration(t *testing.T) {
...
+	}, {
+		name: "new format - label selector",
+		config: &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: system.Namespace(),
+				Name:      IstioConfigName,
+			},
+			Data: map[string]string{
+				"external-gateways": `
+                  - namespace: "namespace"
```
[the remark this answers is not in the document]
pastequo: I can't used tabs for this data, so the global look depends on how tabs are displayed. I add a small helper, let me know if it's better

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -445,3 +469,101 @@ func UpdateGateway(gateway *v1beta1.Gateway, want []*istiov1beta1.Server, existi
...
+		for _, gtw := range gateways[visibility] {
+			ret[visibility] = ret[visibility].Insert(gtw.QualifiedName())
+		}
+	}
+
+	return ret, nil
+}
+
+// gatewaysFromContext get gateways relevant to this ingress from context.
+func gatewaysFromContext(ctx context.Context, ing *v1alpha1.Ingress) (map[v1alpha1.IngressVisibility][]config.Gateway, error) {
```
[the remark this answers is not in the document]
pastequo: Is it safe to assume that the labels are propagated over all owned resources ?

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -445,3 +469,101 @@ func UpdateGateway(gateway *v1beta1.Gateway, want []*istiov1beta1.Server, existi
 func isPlaceHolderServer(server *istiov1beta1.Server) bool {
 	return cmp.Equal(server, &placeholderServer, protocmp.Transform())
 }
+
+// QualifiedGatewayNamesFromContext get gateway names from context.
+func QualifiedGatewayNamesFromContext(ctx context.Context, ing *v1alpha1.Ingress) (map[v1alpha1.IngressVisibility]sets.Set[string], error) {
```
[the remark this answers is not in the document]
pastequo: That is what I initially though here https://github.com/knative-extensions/net-istio/pull/1218 but it requires additional changes in knative serving repo (see @nak3 comment)

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -395,24 +396,47 @@ func GetNonWildcardIngressTLS(ingressTLS []v1alpha1.IngressTLS, nonWildcardSecre
...
+	parts := strings.SplitN(ingressgateway.ServiceURL, ".", 3)
+	if len(parts) != 3 {
+		return ret, fmt.Errorf("unexpected service URL form: %s", ingressgateway.ServiceURL)
+	}
+
+	ret.Name = parts[0]
+	ret.Namespace = parts[1]
+
+	return ret, nil
+}
```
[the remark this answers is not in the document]
pastequo: (I didn't add this TODO, just did a little refacto around this code)

Then we would need to add service cluster suffix as well

As a user (just my personal opinion), I prefer to write
```yaml
serviceURL: myservice.mynamespace.svc.cluster.local
```
than 
```yaml
serviceName: myService
serviceNamespace: myNamespace
suffix: svc.cluster.local
```

Here is the issue: https://github.com/knative-extensions/net-istio/issues/1280

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -445,3 +469,101 @@ func UpdateGateway(gateway *v1beta1.Gateway, want []*istiov1beta1.Server, existi
...
+
+func PublicGatewayServiceURLFromContext(ctx context.Context, ing *v1alpha1.Ingress) (string, error) {
+	return getGatewayServiceURLFromContext(ctx, ing, v1alpha1.IngressVisibilityExternalIP)
+}
+
+func PrivateGatewayServiceURLFromContext(ctx context.Context, ing *v1alpha1.Ingress) (string, error) {
+	return getGatewayServiceURLFromContext(ctx, ing, v1alpha1.IngressVisibilityClusterLocal)
+}
```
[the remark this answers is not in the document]
pastequo: These method needs to filter the gateways that apply to this specific ingress. This is why it's not in the config package, it would create a circular dependency

Or, as you mentioned, this method takes as param the list of gateways

I think I will put it back in ingress where it used to be

pkg/reconciler/ingress/ingress.go
```diff
@@ -258,8 +262,18 @@ func (r *Reconciler) reconcileIngress(ctx context.Context, ing *v1alpha1.Ingress
 	}
 
 	if ready {
-		publicLbs := getLBStatus(publicGatewayServiceURLFromContext(ctx))
-		privateLbs := getLBStatus(privateGatewayServiceURLFromContext(ctx))
+		publicGateways, err := resources.PublicGatewayServiceURLFromContext(ctx, ing)
```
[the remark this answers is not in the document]
pastequo: defaultGateways only contains the qualified names (as it used to do) but you are definitively right the simplification

I wouldn't be in favor to pass this list of gateway everywhere tho (and change the signature in the resource package into something like "context, ingress, gateways"). Only for the helper right here seems better. WDYT ?

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -445,3 +469,79 @@ func UpdateGateway(gateway *v1beta1.Gateway, want []*istiov1beta1.Server, existi
...
+
+	for _, gtw := range gtws {
+		if gtw.LabelSelector == nil { // default value
+			continue
+		}
+
+		selector, err := metav1.LabelSelectorAsSelector(gtw.LabelSelector)
+		if err != nil {
+			return ret, fmt.Errorf("failed to create selector from gateway (%s) label selector: %w", gtw.QualifiedName(), err)
+		}
```
dprotaso, line 537: We should validate the selectors when validating the configmap so the operator can fix it immediately.

pkg/reconciler/ingress/config/istio.go
```diff
@@ -207,8 +231,12 @@ func parseNewFormat(configMap *corev1.ConfigMap) (*Istio, error) {
...
-		return ret, fmt.Errorf("only one local gateway can be defined: current %q value is %q", localGatewaysKey, localGatewaysStr)
+	if len(ret.DefaultExternalGateways()) > 1 {
+		return ret, fmt.Errorf("only one external gateway with no selector can be defined, here: %v", ret.DefaultExternalGateways())
+	}
+
+	if len(ret.DefaultLocalGateways()) > 1 {
```
dprotaso, line 238: I think we should always ensure the operator configures a default gateway - otherwise there are bunch of error cases we would need to handle in other parts of the code (eg. ingress not matching _any_ gateway)


```suggestion
	if len(ret.DefaultExternalGateways()) != 1  {
		return ret, fmt.Errorf("only one external gateway with no selector can be defined, here: %v", ret.DefaultExternalGateways())
	}

	if len(ret.DefaultLocalGateways()) !== 1 {
```

test/e2e/exposition_test.go
```diff
@@ -0,0 +1,260 @@
...
+	"knative.dev/net-istio/pkg/reconciler/ingress/config"
+	"knative.dev/networking/pkg/apis/networking"
+	"knative.dev/networking/pkg/apis/networking/v1alpha1"
+	"knative.dev/networking/test"
+	"knative.dev/networking/test/conformance/ingress"
+	"knative.dev/pkg/system"
+)
+
+func TestExposition(t *testing.T) {
+	clients := Setup(t)
```
dprotaso, line 38: We probably want to skip this test when doing mesh testing

If the env var `MESH` is set you could skip this test

pkg/reconciler/ingress/config/istio.go
dprotaso: You might have to update your e2e test

dprotaso:
This is looking great. Some minor stuff left. Would be good to get reviews from from the other approvers as a final pass

/assign @skonto @ReToCode

ReToCode (commented):
Looks good to me. @skonto anything to add?

ReToCode:
@pastequo want to add some release notes to the PR description?

pastequo:
I added something in the PR description, let me know if I should be more specific, illustrative or anything

ReToCode:
> I added something in the PR description, let me know if I should be more specific, illustrative or anything

That's perfect, I just on-lined it to not break the generated release notes.

skonto:
I think we should document this feature. Is this something we plan to do?

pkg/reconciler/ingress/resources/gateway.go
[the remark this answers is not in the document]
skonto: So we keep the 1-1 for ingress gateways and istio gateways for now?

test/e2e/exposition_test.go
```diff
@@ -0,0 +1,251 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2020 The Knative Authors
```
skonto, line 5: s/2020/2024?

skonto:
With this work does that mean we need to extend https://github.com/knative/operator/blob/main/pkg/reconciler/knativeserving/ingress/istio.go#L58 somehow?

pkg/reconciler/ingress/config/istio.go
```diff
@@ -103,35 +105,62 @@ func (g Gateway) Validate() error {
...
 		if err := gtw.Validate(); err != nil {
-			return fmt.Errorf("invalid local gateway: %w", err)
+			return fmt.Errorf("invalid local gateway %s: %w", gtw.QualifiedName(), err)
 		}
 	}
 
 	return nil
 }
 
+func (i Istio) DefaultExternalGateways() []Gateway {
```
skonto, line 142: Docs for public method are missing. Same for `DefaultLocalGateways()`

pkg/reconciler/ingress/config/istio_test.go
```diff
@@ -233,14 +234,14 @@ func TestGatewayConfiguration(t *testing.T) {
...
 			Data: map[string]string{
-				"external-gateways": `
-                  - namespace: "namespace1"
-                    name: "gateway1"
-                    service:  "istio-ingressbackroad.istio-system.svc.cluster.local"`,
-				"local-gateways": `
-                  - namespace: "namespace2"
-                    name: "gateway2"
-                    service:  "istio-local-gateway.istio-system.svc.cluster.local"`,
+				"external-gateways": replaceTabs(`
```
skonto, line 237: Should not we have a test for defaults to:

```
{
		name: "new format alone defaults",
		wantIstio: &Istio{
			IngressGateways: []Gateway{{
				Namespace:  system.Namespace(),
				Name:       "knative-ingress-gateway",
				ServiceURL: "istio-ingressgateway.istio-system.svc.cluster.local",
			}},
			LocalGateways: []Gateway{{
				Namespace:  system.Namespace(),
				Name:       "knative-local-gateway",
				ServiceURL: "knative-local-gateway.istio-system.svc.cluster.local",
			}},
		},
		config: &corev1.ConfigMap{
			ObjectMeta: metav1.ObjectMeta{
				Namespace: system.Namespace(),
				Name:      IstioConfigName,
			},
		},
	}
```

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -146,12 +147,12 @@ func MakeExternalIngressGateways(ctx context.Context, ing *v1alpha1.Ingress, ser
 
 // MakeWildcardTLSGateways creates gateways that only contain TLS server with wildcard hosts based on the wildcard secret information.
 // For each public ingress service, we will create a list of Gateways. Each Gateway of the list corresponds to a wildcard cert secret.
-func MakeWildcardTLSGateways(ctx context.Context, originWildcardSecrets map[string]*corev1.Secret,
+func MakeWildcardTLSGateways(ctx context.Context, ing *v1alpha1.Ingress, originWildcardSecrets map[string]*corev1.Secret,
```
skonto, line 150: Since we are creating the gateways based on the ingress info as well, we probably need to update the description to reflect that.

pkg/reconciler/ingress/resources/gateway.go
```diff
@@ -395,24 +396,47 @@ func GetNonWildcardIngressTLS(ingressTLS []v1alpha1.IngressTLS, nonWildcardSecre
...
-	for i, ingressgateway := range cfg.IngressGateways {
-		parts := strings.SplitN(ingressgateway.ServiceURL, ".", 3)
-		if len(parts) != 3 {
-			return nil, fmt.Errorf("unexpected service URL form: %s", ingressgateway.ServiceURL)
-		}
-		nameNamespaces[i] = metav1.ObjectMeta{
-			Name:      parts[0],
-			Namespace: parts[1],
+// GetIngressGatewaySvcNameNamespaces gets the Istio ingress namespaces from ConfigMap for gateways that should expose the service.
+func GetIngressGatewaySvcNameNamespaces(ctx context.Context, obj kmeta.Accessor) ([]metav1.ObjectMeta, error) {
```
skonto, line 400: It seems to me that wherever we have a kmeta.Accessor we could replace it with `ing *v1alpha1.Ingress`, correct?

test/e2e/exposition_test.go
```diff
@@ -0,0 +1,251 @@
...
+			cm.Data = oldConfig.Data
+
+			if _, err := clients.KubeClient.CoreV1().ConfigMaps(namespace).Update(context.Background(), cm, metav1.UpdateOptions{}); err != nil {
+				t.Fatalf("Failed to restore configmap %s/%s: %v", namespace, config.IstioConfigName, err)
+			}
+		})
+	}
+}
+
+func replaceTabs(s string) string {
```
skonto, line 249: We could have as an alternative:

```
func toYamlNewFormatGateways(data []Gateway) string {
	bytes, _ := yaml.Marshal(data)
	return string(bytes)
}

"external-gateways": toYamlNewFormatGateways([]Gateway{
  {
    Namespace:  "unused",
    Name:       "unused",
    ServiceURL: "default.default.svc.cluster.local",
  },
  {
    Namespace:  "namespace",
    Name:       "gateway",
    ServiceURL: "istio-gateway.istio-system.svc.cluster.local",
    LabelSelector: &metav1.LabelSelector{
      MatchExpressions: []metav1.LabelSelectorRequirement{
        {
          Key:      "key",
          Operator: metav1.LabelSelectorOpIn,
          Values:   []string{"value"},
        },
      },
    },
  }},
),
```

pkg/reconciler/ingress/config/istio_test.go
pastequo: There is the very first one

```golang
	{
		name: "gateway configuration with no network input",
		wantIstio: &Istio{
			IngressGateways: defaultIngressGateways(),
			LocalGateways:   defaultLocalGateways(),
		},
		config: &corev1.ConfigMap{
			ObjectMeta: metav1.ObjectMeta{
				Namespace: system.Namespace(),
				Name:      IstioConfigName,
			},
		},
	}, 
```

pastequo: (default doesn't matter if it's the old format or the new one)

pkg/reconciler/ingress/resources/gateway.go
pastequo: lmk if it's more clear

pkg/reconciler/ingress/resources/gateway.go
pastequo: True. I used to have `ing *v1alpha1.Ingress` & @dprotaso made me change for this. Which is imho better to depends on a small interface rather than an implementation

See https://github.com/knative-extensions/net-istio/pull/1250#discussion_r1496262747

test/e2e/exposition_test.go
pastequo: I think I rather have a string that reflects exactly what a user would have written, rather than generate it

skonto: For the scenarios you want to check just the user syntax yes, for the rest where you want to run test cases based on valid content it is safer. Anyway no strong preference.

pkg/reconciler/ingress/resources/gateway.go
skonto: It depends generalizing for one case it is still weird but maybe a good fir for the GetLabels method.

pastequo:
> I think we should document this feature. Is this something we plan to do?

Do you mean aside from the google doc ? https://docs.google.com/document/d/12X1-9nhjAhpf-Wlt3RbdoMbvx31zFBva/edit#heading=h.gjdgxs

Something here https://github.com/knative/docs ? Or somewhere else ?

pkg/reconciler/ingress/config/istio_test.go
skonto: ok that's good then.

pkg/reconciler/ingress/resources/gateway.go
skonto: Maybe rephrase to "Gateways generated are based on the related ingress being reconciled".

pastequo:
Issue created: https://github.com/knative/docs/issues/5905

dprotaso:
/approve 

Leaving lgtm for stavros

skonto:
@pastequo could you rebase?

dprotaso:
🎉 thanks @pastequo for your help on landing this

Issue #1247, opened 2024-01-25
New configuration format


pastequo:
This Issue comes from this design document: https://docs.google.com/document/d/12X1-9nhjAhpf-Wlt3RbdoMbvx31zFBva/edit#heading=h.gjdgxs

The proposal is to make the configuration more flexible, more extensible. This issue focus only on the new configuration format, without any filtering capabilities (that will come in another issue)

See the design document for more information

nguyenthai0107:
Hello @pastequo .
Is there a way to implement with Nginx Ingress ?.
Im believe that a lot of company which was built and using technology and system base on Nginx Ingress ( without supported network layer such as Contour, Kourier, Istio). It will more options, more flexible if this was implemented in configmap.
Thank you.

dprotaso:
@nguyenthai0107 nginx support would come through our use of the Gateway API - we have a plugin here but we haven't done any testing with nginx.

https://github.com/knative-extensions/net-gateway-api

Curious do you require the need to select different gateways?

PR #1257, opened 2024-02-08
docs: Add documentation about the new config format (#1247)

pastequo:
# Changes

- Documentation about the new configuration format

/kind <kind>

Fixes #

### DIFF
--- a/config/400-config-istio.yaml
+++ b/config/400-config-istio.yaml
@@ -65,13 +65,6 @@ data:
     # Please use the new configuration format `external-gateways` for future compatibility.
     # This configuration format is ignored if either `external-gateways` or `local-gateways` is defined.
     gateway.knative-serving.knative-ingress-gateway: "istio-ingressgateway.istio-system.svc.cluster.local"
-    #
-    #
-    # If no values are specified for the external traffic, it will default to:
-    # external-gateways: |
-    #  - name: knative-ingress-gateway
-    #    namespace: {{serving_system_namespace}}
-    #    service: istio-ingressgateway.istio-system.{{cluster_domain}}
 
 
     # local-gateways defines a cluster local gateway to allow pods outside of the mesh to access
@@ -107,10 +100,3 @@ data:
     # Please use the new configuration format `local-gateways` for future compatibility.
     # This configuration format is ignored if either `external-gateways` or `local-gateways` is defined.
     local-gateway.knative-serving.knative-local-gateway: "knative-local-gateway.istio-system.svc.cluster.local"
-    #
-    #
-    # If no values are specified for the local traffic, it will default to:
-    # local-gateways: |
-    #  - name: knative-local-gateway
-    #    namespace: {{serving_system_namespace}}
-    #    service: knative-local-gateway.istio-system.{{cluster_domain}}
--- a/config/400-config-istio.yaml
+++ b/config/400-config-istio.yaml
@@ -41,9 +41,10 @@ data:
     # this example block and unindented to be in the data block
     # to actually change the configuration.
 
-    # A gateway and Istio service to serve external traffic.
-    #
-    # The configuration format should be
+
+    # external-gateways defines a gateway and Istio service to serve external traffic.
+    # It is the new and preferred way to define the configuration.
+    # The format is as follow:
     # ```
     # external-gateways: |
     #   - name: {{gateway_name}}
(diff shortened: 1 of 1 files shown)


config/400-config-istio.yaml
```diff
@@ -56,10 +83,44 @@ data:
...
     # The configuration format should be `local-gateway.{{local_gateway_namespace}}.
     # {{local_gateway_name}}: "{{cluster_local_gateway_name}}.
     # {{cluster_local_gateway_namespace}}.svc.cluster.local"`. The
     # {{local_gateway_namespace}} is optional; when it is omitted, the system
     # will search for the local gateway in the serving system namespace
     # `knative-serving`
     local-gateway.knative-serving.knative-local-gateway: "knative-local-gateway.istio-system.svc.cluster.local"
+
+
+    # If no values are specified for the external traffic, it will default to:
```
ReToCode, line 116: I'd probably keep the blocks together. Now this keys are kind of at the "old format" title.

config/400-config-istio.yaml
```diff
@@ -56,10 +83,44 @@ data:
     # sidecar injection (like Knative's istio-ci-no-mesh.yaml).  Since every pod
     # is outside of the service mesh in that case, a cluster-local  service
     # will need to be exposed to a cluster-local gateway to be accessible.
+    #
+    ##############
+    # NEW FORMAT #
+    ##############
```
ReToCode, line 89: Seems a bit verbose. I think it is fine to just point from the old key to the new one, e.g. as in https://github.com/knative/networking/blob/main/config/config-network.yaml#L117. We don't need to say deprecated, but something like you have above "should be preferred.".

config/400-config-istio.yaml
```diff
@@ -41,13 +41,40 @@ data:
     # this example block and unindented to be in the data block
     # to actually change the configuration.
 
+
+    # Documentation below explicits both new and old format.
```
ReToCode, line 45: Can you integrate that with the docs on the keys? I'm not sure if anybody reads these lines (people tend to look at the keys directly).

config/400-config-istio.yaml
```diff
@@ -42,11 +42,30 @@ data:
     # to actually change the configuration.
 
     # A gateway and Istio service to serve external traffic.
```
ReToCode, line 44:
```suggestion
    # external-gateways defines a gateway and Istio service to serve external traffic. It is the new and preferred way to define the configuration. The format is as the following example:
```

config/400-config-istio.yaml
```diff
@@ -42,11 +42,30 @@ data:
...
+    #     namespace: {{gateway_namespace}}
+    #     service: {{ingress_name}}.{{ingress_namespace}}.svc.cluster.local
+    # ```
+    # name, namespace & service are mandatory and can't be empty.
+    external-gateways: |
+      - name: knative-ingress-gateway
+        namespace: knative-serving
+        service: istio-ingressgateway.istio-system.svc.cluster.local
+    #
+    # Format `gateway.{{gateway_namespace}}.{{gateway_name}}: "{{ingress_name}}.{{ingress_namespace}}.svc.cluster.local"`
```
ReToCode, line 59:
```suggestion
    #  
    # The old format has the following pattern
    # `gateway.{{gateway_namespace}}.{{gateway_name}}: "{{ingress_name}}.{{ingress_namespace}}.svc.cluster.local"`
    #  Please use the new configuration format `external-gateways` for future compatibility.
    # This configuration format is ignored if either `external-gateways` or `local-gateways` is defined.
```

ReToCode (commented):
I think this is better, but still a bit confusing. Can we have a description that looks similar in each block with spacing between the blocks?

I added two inline suggestions to showcase what I mean.

pastequo:
It's not crystal clear for me what you have in mind. I included your suggestions, it's probably not enough. Feel free to add your commits on top of it.

config/400-config-istio.yaml
```diff
@@ -41,25 +41,76 @@ data:
...
+    # The old format has the following pattern:
+    # ```
+    # gateway.{{gateway_namespace}}.{{gateway_name}}: "{{ingress_name}}.{{ingress_namespace}}.svc.cluster.local"
+    # ```
+    # Please use the new configuration format `external-gateways` for future compatibility.
+    # This configuration format is ignored if either `external-gateways` or `local-gateways` is defined.
     gateway.knative-serving.knative-ingress-gateway: "istio-ingressgateway.istio-system.svc.cluster.local"
+    #
+    #
+    # If no values are specified for the external traffic, it will default to:
```
ReToCode, line 70: Looks very good now. I think this block is not necessary. We already have more or less the "default" at line 55. Maybe we can use {{cluster_domain}} at line 52 to make this part also clear.

config/400-config-istio.yaml
```diff
@@ -41,25 +41,76 @@ data:
...
+    # local-gateway.{{local_gateway_namespace}}.{{local_gateway_name}}:
+    # "{{cluster_local_gateway_name}}.{{cluster_local_gateway_namespace}}.svc.cluster.local"
+    # ```
+    # Please use the new configuration format `local-gateways` for future compatibility.
+    # This configuration format is ignored if either `external-gateways` or `local-gateways` is defined.
     local-gateway.knative-serving.knative-local-gateway: "knative-local-gateway.istio-system.svc.cluster.local"
+    #
+    #
+    # If no values are specified for the local traffic, it will default to:
+    # local-gateways: |
```
ReToCode, line 113: Same as above.

config/400-config-istio.yaml
ReToCode: Or, if we want to keep the block, it should be above the old format. It seems a bit odd to go back to the new key after the old key.

pastequo: Agree, it was done that way because it's the default value if none of the new and old format is defined.
I think I will just remove it, it's confusing wherever it is.

pastequo: Done

ReToCode (approved):
Thanks for bearing with me. I think now its very easy to follow and understand.

/lgtm
/approve

PR #1325, opened 2024-05-28
fix: Fixed liveness periodseconds to 10 so that crashloopback off doe…

mukulgit123:
# Changes

Fixed liveness periodseconds to 10 so that crashloopback off doesn't happen which causes expired lease to get stuck.

Which is already fixed in knative-serving with https://github.com/knative/serving/issues/15255

/kind bug

Fixes https://github.com/knative/serving/issues/15255

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/config/500-webhook-deployment.yaml
+++ b/config/500-webhook-deployment.yaml
@@ -87,7 +87,7 @@ spec:
             port: 8443
           failureThreshold: 3
         livenessProbe:
-          periodSeconds: 1
+          periodSeconds: 10
           httpGet:
             scheme: HTTPS
             port: 8443


ReToCode:
Hm probably introducing startup-probes is a better approach. This would also set the later lifeness-check to only every 10 seconds. This massively increases the time until the pod would be considered unhealthy: 10*6 = 60 seconds vs 10*1s.
WDYT @dprotaso @skonto ?

dprotaso:
/lgtm
/approve

> Hm probably introducing startup-probes is a better approach. 

Want to do some follow up PRs? Let's merge this in for now.

ReToCode:
@dprotaso  no concerns about?

> This massively increases the time until the pod would be considered unhealthy: 10 * 6 = 60 seconds vs 10 * 1s.

dprotaso:
> This massively increases the time until the pod would be considered unhealthy: 10 * 6 = 60 seconds vs 10 * 1s.

Unsure - because we've had now two scenarios in the past where the liveness kicks and restarts the pod before the webhook is ready.

Maybe our liveness probes are wrong and shouldn't be dependent on HTTPS.

This PR just fixes the stale lease issue - and we can make adjustments before v1.15 is out.

PR #1329, opened 2024-05-31
Add missing common labels on kcert

ReToCode:
# Changes
- Add missing common labels on kcert
- Without these, the operator does not properly select these YAMLs

/assign @skonto 

/cherry-pick release-1.14


skonto:
Are there any other resources that need this or we are good?

ReToCode:
> Are there any other resources that need this or we are good?

I think the rest is fine, I added this YAML and just missed the labels

PR #1332, opened 2024-06-06
use cgr.dev/chainguard/static:latest which is nonroot

dprotaso:
Part of https://github.com/knative/serving/pull/15310


ReToCode (approved):
/lgtm
/approve

PR #1343, opened 2024-07-10
Bump Istio and simplify update script

dprotaso:
Now we just need to bump a number in ./hack/update-dep.sh
then run the script to update our golang dependencies
and the generated manifests

Also it seems like the `istio generate manifest` command
outputs stable documents so we don't need our sorting logic
anymore


ReToCode (approved):
Nice, thanks @dprotaso 

/lgtm
/approve

PR #1369, opened 2024-09-24
Use istio.sidecar.inject label instead of annotation

ReToCode:
Partially https://github.com/knative/serving/issues/15521

## Proposed Changes
* Use istio.sidecar.inject label instead of annotation

/assign @skonto

### DIFF
--- a/config/500-controller.yaml
+++ b/config/500-controller.yaml
@@ -28,12 +28,11 @@ spec:
       app: net-istio-controller
   template:
     metadata:
-      annotations:
+      labels:
         # This must be outside of the mesh to probe the gateways.
         # NOTE: this is allowed here and not elsewhere because
         # this is the Istio controller, and so it may be Istio-aware.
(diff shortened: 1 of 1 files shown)


skonto:
It seems Istio specifies both https://github.com/knative-extensions/net-istio/blob/main/third_party/istio-latest/istio-ci-no-mesh/istio.yaml#L14297. Should we keep it that why? :thinking:

dprotaso:
> It seems Istio specifies both https://github.com/knative-extensions/net-istio/blob/main/third_party/istio-latest/istio-ci-no-mesh/istio.yaml#L14297. Should we keep it that why? 🤔 Are there any Istio versions that this makes a difference?

That's a good point - which istio version added support for this in labels?

ReToCode:
Support was added in 1.11 (https://istio.io/latest/news/releases/1.11.x/announcing-1.11/#updates-to-revision-and-tag-deployments). So I think this should be fine?

PR #1395, opened 2025-01-14
Expose PROBE_MAX_RETRY_DELAY_SECONDS configuration with default 30

SaschaSchwarze0:
# Changes

Changing this to a lower value improves the time it takes to provision a new or updated KService on large clusters. See https://github.com/knative/networking/pull/1001 for more information.

/kind performance

**Release Note**

```release-note
You can now change the environment variable PROBE_MAX_RETRY_DELAY_SECONDS to a lower value (default is 30) to improve provisioning time of KServices in larger clusters.
```

**Docs**

### DIFF
--- a/config/500-controller.yaml
+++ b/config/500-controller.yaml
@@ -69,6 +69,9 @@ spec:
         - name: METRICS_DOMAIN
           value: knative.dev/net-istio
 
+        - name: PROBE_MAX_RETRY_DELAY_SECONDS
+          value: "30"
+
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true


config/500-controller.yaml
```diff
@@ -69,6 +69,11 @@ spec:
         - name: METRICS_DOMAIN
           value: knative.dev/net-istio
 
+        # Improve provisioning time of new or updated kservice in large clusters
+        # see https://github.com/knative/networking/pull/1001
+        - name: PROBE_MAX_RETRY_DELAY_SECONDS
+          value: "5"
```
skonto, line 75: I would rather keep the default here and a note in docs or here as a comment about large clusters.

SaschaSchwarze0: I am okay to change to the default here. Before I start seaching, can you point me to a good place where to document this?

What I do not see is the higher traffic issue. In a cluster with KServices that are really used, no matter if it is small or large, the amount of probe requests relative to all requests will be low if not irrelevant.

SaschaSchwarze0:
No, nothing noticable. As mentioned earlier, in our ingress gateway pods, the probe requests play no role at all compared to the overall throughput. Net-istio controller we had not specifically looked at as that controller was and is generally needing a lot fewer compute resources compared to for example the serving controller or the autoscaler.

config/500-controller.yaml
skonto: > the amount of probe requests relative to all requests
I depends on the traffic I guess. What I mean is that I want to avoid something observable from other users.

skonto:
@dprotaso should we keep it to 30 and users can adjust as needed? Are we ok to stamp this?

dprotaso:
I believe this change is a noop correct?

The default max retry was 30s - https://github.com/knative/networking/pull/1001/files#diff-1deea3647518183ee4915dd3c10ccfc7bb60e1ed860673c99eefaf5cfeefaa6bR63

dprotaso:
I'm going to close this out as I don't think we need this explicitly in the config

PR #1410, opened 2025-02-14
allow removing the finalizer if a gateway is missing

dprotaso:
Fixes: https://github.com/knative-extensions/net-istio/issues/1407

/assign @skonto


skonto:
@dprotaso does that mean that if someone has a configmap that is not aligned with the current ingress we allow to delete?

dprotaso:
> does that mean that if someone has a config that is not aligned with the currently deployed ingresses on the cluster we allow skip the "not found" error during deletion?

yes exactly

PR #1415, opened 2025-02-19
[release-1.16] Move codegen script to new kube_codegen.sh

dsimansk:
# Changes

- [release-1.16] Move codegen script to new kube_codegen.sh

PTAL, this is a usual codegen refactor to make sure `release-1.16` codegen scripts work with kube_codegen.sh stuff. Later release branch and main is OK. 
 
/cc @skonto @dprotaso


dprotaso:
Yeah it seems it’s broken and we need productivity to look into it

PR #1416, opened 2025-02-25
reading vs destination host from istio configmap

yuehaii:
# Changes

The virtual service destination host should not read from resolv.conf. It should be same with envoy cluster which generated by CDS discovery service. 
The pilot CDS service discover the k8s services and add a cluster domain suffix. This cluster domain is configured in istio config map. Even change this cluster domain setting, it will not cause resolv.conf record been refreshed.
Most istio user will not change this default cluster domain, and k8s will also add the default svc.cluster.local into each pod resolv.conf file. So reading virtual service destination host from resolv.conf works most of time. 
But if the istio cluster domain been changed, or k8s cloud provider changed the resolv.conf, it will cause the virtual service routing breaking with cluster not found.

/kind bug


dprotaso:
> The virtual service destination host should not read from resolv.conf.

These host names are generated by the knative serving route controller when we create the Knative Ingress resource.

It is using resolve.conf
https://github.com/knative/pkg/blob/0372c089c78f40ce1009f6d20ccab75205dea143/network/domain.go#L46

> But if the istio cluster domain been changed, or k8s cloud provider changed the resolv.conf, it will cause the virtual service routing breaking with cluster not found.

We don't expect the cluster domain to change _after_ the cluster has been created.

If the resolve.conf approach doesn't work we have an override env var available - see: https://github.com/knative/pkg/blob/main/network/domain.go#L30

You'll want to set this on the serving `controller` deployment.

I'm curious though - why would the cluster domain change _after_ creation?

yuehaii:
> > The virtual service destination host should not read from resolv.conf.
> 
> These host names are generated by the knative serving route controller when we create the Knative Ingress resource.
> 
> It is using resolve.conf https://github.com/knative/pkg/blob/0372c089c78f40ce1009f6d20ccab75205dea143/network/domain.go#L46
> 
> > But if the istio cluster domain been changed, or k8s cloud provider changed the resolv.conf, it will cause the virtual service routing breaking with cluster not found.
> 
> We don't expect the cluster domain to change _after_ the cluster has been created.
> 
> If the resolve.conf approach doesn't work we have an override env var available - see: https://github.com/knative/pkg/blob/main/network/domain.go#L30
> 
> You'll want to set this on the serving `controller` deployment.
> 
> I'm curious though - why would the cluster domain change _after_ creation?

It is not changing the domain after creation. When we install istio with helm, there is a configuration value can specify the clusterDomain:  https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/values.yaml#L311 
During investigation and experiment, we verified that this value can be changed after creation.  So when user install istio, the cluster domain can be configured and cause such issue. 

The other scenario to cause such issue is changing that resolv.conf. E.g. cloud provider change resolv.conf file, and append cluster id to "cluster.local"

dprotaso:
> E.g. cloud provider change resolv.conf file, and append cluster id to "cluster.local"

Interesting what implementations do this?

dprotaso:
Also if providers do that - wouldn't you need to update the istio config map with the new cluster local domain?

It just seems like an odd edge case that providers would break these workloads

dprotaso:
I'm going to close this out - I don't think we want to support this edge case. I would expect the cluster domain to be stable prior to installing Knative and Istio.

Also the solution here is istio specific. In general if we wanted to support this edge case scenario we would have to make this work more generally for other net-* ingress plugins.

PR #1426, opened 2025-04-17
increase test coverage

dprotaso:


dsimansk (changes requested):
Missing license headers in new files.

PR #1436, opened 2025-06-10
bump istio to v1.26.1

dprotaso:


dprotaso:
/retest

Cluster didn't start

> : All cluster resources were brought up, but: only 8 nodes out of 9 have registered; cluster may be unhealthy.\n"
test/e2e-tests.sh failed: 1

PR #1454, opened 2025-08-21
add request mirroring to net-istio

shraykay:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/reconciler/ingress/resources/virtual_service.go
+++ b/pkg/reconciler/ingress/resources/virtual_service.go
@@ -18,6 +18,7 @@ package resources
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -133,7 +134,7 @@ func makeVirtualServiceSpec(ing *v1alpha1.Ingress, gateways map[v1alpha1.Ingress
 			p := rule.HTTP.Paths[i]
 			hosts := hosts.Intersection(sets.New(rule.Hosts...))
 			if hosts.Len() != 0 {
-				http := makeVirtualServiceRoute(hosts, &p, gateways, rule.Visibility)
+				http := makeVirtualServiceRoute(hosts, &p, gateways, rule.Visibility, ing)
 				// Add all the Gateways that exist inside the http.match section of
 				// the VirtualService.
 				// This ensures that we are only using the Gateways that actually appear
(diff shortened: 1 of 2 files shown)


dprotaso:
Do you have examples of the types of 'host' you mirror to?

I'm wondering if there's any security issues here - eg. sending traffic to an arbitrary host.

dprotaso:
Secondly, one caveat is we track revision reachability when a route is pointing to it.

This allows us to optimize etc min-scale etc (eg. when no route is pointed to it then we scale to zero instead of min-scale).

Because of this it seems like the 'safe' way to mirror would be to another Knative Service and not a Revision/Host.

shraykay:
> Do you have examples of the types of 'host' you mirror to?
> 
> I'm wondering if there's any security issues here - eg. sending traffic to an arbitrary host.


@dprotaso

In the kserve world we would like to be able to spin up two `InferenceService` a primary and a secondary - each of which creates a Knative service for the predictor and one for the transformer. When we're testing the performance of a new InferenceService, we'd like to mirror traffic from the primary production transformer to a secondary transformer and measure the throughput metrics. In this case, it would be:

```
kubectl get svc search-second-pass-predictor
NAME                           TYPE           CLUSTER-IP   EXTERNAL-IP                                             PORT(S)   AGE
search-second-pass-predictor   ExternalName   <none>       knative-local-gateway.istio-ingress.svc.cluster.local   80/TCP    27d
```

so on the primary production tranformer knative service:
```
networking.knative.dev/request-mirror: search-second-pass-predictor.kserve.svc.cluster.local
networking.knative.dev/request-mirror-port: 80
networking.knative.dev/request-mirror-percentage: 100
```

Are the security concerns relevant to the net-istio project? From what I understand this is a envoy configuration that is being manipulated.

dprotaso:
> Are the security concerns relevant to the net-istio project? 

This might be more theoretical for example if someone sets CNI rules/Istio policy rules that disallow a set of Knative services in namespace A can't communicate to namespace B. But setting up the mirror'ing _could_ allow you to bypass that since it's coming from the ingress.  Though I'd imagine you could have strict policy there. Also the activator is a shared component 🤷 . @evankanderson can keep me honest.

The main concern is we have a labeller controller that's marking revisions as 'reachable' that controls whether revisions scale to zero. In order to use your suggested mirroring feature you need to ensure a revision is pointed to by a route which might not be obvious.

I'm wondering if it's better to create two InferenceServices and then use a one off VS host to mirror traffic between them

shraykay:
that is fair, thank you!

evankanderson:
I'm trying to understand how this feature is intended to work.

Would it be reasonable to restrict the usage (for now) to Kubernetes services in the same namespace?  That seems to be a common Kubernetes security boundary, and would reduce the concerns about a developer in namespace A being able to "smuggle" requests to namespace B using an unexpected VirtualService with different security settings.

PR #1459, opened 2025-09-10
Fix Server TLS settings so it works with Istio 1.27

dprotaso:
A user reported on slack that net-istio wasn't working with Istio Ambient 1.27

```
admission webhook "validation.istio.io" denied the request: configuration is invalid: only one of credential_name, credential_names, server_certificate/private_key, tls_certificates should be specified
```

Talking to Istio folks it seems like we have been setting extra properties that we don't need.


dprotaso:
Testing downstream - https://github.com/knative/serving/pull/16077

dprotaso:
ok downstream looks good - just the know flakes flaking

/assign @dsimansk @nader-ziada @skonto

PR #1460, opened 2025-09-10
Fix Istio TLS config conflict by removing ServerCertificate/PrivateKey fields

rajatvig:
Removes conflicting ServerCertificate and PrivateKey fields from TLS server settings that cause Istio admission webhook validation errors.

The Istio webhook fails with: "only one of credential_name, credential_names, server_certificate/private_key, tls_certificates should be specified"

Uses only CredentialName for TLS configuration.

- Remove ServerCertificate and PrivateKey fields from TLS configurations  
- Update tests to match new format


dprotaso:
This is a dupe of  https://github.com/knative-extensions/net-istio/pull/1459

PR #1482, opened 2025-11-04
Bump knative deps + fix build

dprotaso:
- **run ./hack/update-deps.sh --upgrade**
- **fix build**


dprotaso:
/assign @nader-ziada 

Had to bump this manually due to the build error

PR #1491, opened 2025-11-26
bump Istio to v1.28.0

dprotaso:


dprotaso:
/retest

odd we're not actually testing the gateway api path

PR #1507, opened 2026-02-11
bump k8s kind

dprotaso:


linkvt (commented):
Active support of 1.33 ends on the current release date of knative 1.22 - no need to test for it then if we don't support it for the release.

/lgtm

PR #1517, opened 2026-02-24
Allow direct ingress to knative services.

### LINKED ISSUE #1003: Provide delegate virtual service to simplify ingress / reduce need for cluster-local-gateway
Originally mentioned here: https://knative.slack.com/archives/C012AK2FPK7/p1660149435405279

One overriding concern I have that limits my knative deployments is the pseudo requirement to pass through so many proxies due to the VirtualService config.  I have to pass the traffic from the ingress gateway to the cluster local gateway then to the activator/service.   It would be awesome if we can reconfigure the knative generated VirtualService to skip the need for the cluster local gateway.    

 The -mesh and -ingress VSs seem identical except for the gateway, any reason this isn't one VS with two gateways?  In any case, if the only VS that needed updating was a delegate VS we could wire that right in from [our] top level VS and skip that cluster local gateway.   That same delegate VS could be used in the existing -ingress and -mesh VS definitions.

As an example, I'll use a VS to delegate a single route to a ksvc.

```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: example
spec:
  hosts:
  - api.example.com
  gateways:
    - "ops/ingress"
  http:
    - name: "something"
       match:
       - uri:
           prefix: "/something"
       delegate:
         name: something
         namespace: this
   - name: someksvc
      match:
       - uri:
           prefix: "/somethingelse"
      rewrite:
        # Rewrite the original host header to the host header of  service
        # in order to redirect requests to  service.
        authority: ksvc-name.this.svc.cluster.local
      route:
        - destination:
            host: cluster-local-gateway.istio-system.svc.cluster.local
            port:
              number: 80
```

if Knative would add a VirtualService that is delegatable, the ingress proxy could communicate directly with activator/service without the need of the cluster-local-gateway and the extra hops, the extra network $$ and the extra configuration madness.

Every ksvc gets two VirtualServices.
```
someksvc-ingress               ["knative-serving/cluster-local-gateway"]   ["someksvc.this","someksvc.this.svc","someksvc.this.svc.cluster.local"]                                       7d2h
someksvc-mesh                  ["mesh"]                                    ["someksvc.this","someksvc.this.svc","someksvc.this.svc.cluster.local"]                                       7d2h
```
Knative appears to edit both of those VSs for every change.  This services can not be used for delegates because they contain host names.   If a delegate VS is added with no hosts, we can delegate a route and have the ingress gateway connect to the service directly without the hop through the cluster local gateway.   That same delegate VS could then be added as a route for the -ingress and -mesh gateways and only THAT updated every time knative needs to swap between a service and activator.

treyhyde:
Fixes https://github.com/knative-extensions/net-istio/issues/1003

# Changes

- 🎁 Allow direct ingress via delegate virtual services
-
-

/kind enhancement

Fixes #1003

Knative requires you to use an intermediate hop to a cluster local gateway.

Ingress VS (rewrite authority) -> cluster local gateways -> activator (optional) -> service

This change allows you to delegate your route directly to the service  / activator

Ingress VS -> activator (optional) -> service

 
**Release Note**

```release-note
Creates a delegate VirtualService for every knative service to allow for direct ingress to a service without adding additional hops via a cluster-local-gateway

```

**Docs**

```docs

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@
 # Temporary output of build tools
 bazel-*
 *.out
+
+# binaries
+bin/
--- a/config/400-config-istio.yaml
+++ b/config/400-config-istio.yaml
@@ -119,3 +119,9 @@ data:
     # Please use the new configuration format `local-gateways` for future compatibility.
     # This configuration will raise an error if either `external-gateways` or `local-gateways` is defined.
     local-gateway.knative-serving.knative-local-gateway: "knative-local-gateway.istio-system.svc.cluster.local"
+
+    # enable-delegate-virtual-service controls whether a delegate VirtualService
+    # is created for each Ingress. When enabled, a VirtualService with no hosts
+    # or gateways is created, allowing other VirtualServices to delegate routing
+    # directly to Knative services. Disabled by default.
+    enable-delegate-virtual-service: "false"
--- a/pkg/reconciler/ingress/config/istio.go
+++ b/pkg/reconciler/ingress/config/istio.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
@@ -58,6 +59,9 @@ const (
 
 	// IstioNamespace is the namespace containing Istio
 	IstioNamespace = "istio-system"
+
+	// EnableDelegateVirtualServiceKey is the configmap key to enable delegate VirtualService creation.
+	EnableDelegateVirtualServiceKey = "enable-delegate-virtual-service"
 )
 
 func defaultIngressGateways() []Gateway {
@@ -121,6 +125,10 @@ type Istio struct {
 
 	// LocalGateways specifies the gateway urls for public & private Ingress.
 	LocalGateways []Gateway
+
+	// EnableDelegateVirtualService enables the creation of delegate VirtualServices
+	// that allow direct ingress to Knative services.
+	EnableDelegateVirtualService bool
 }
 
 func (i Istio) Validate() error {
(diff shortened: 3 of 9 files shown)


dprotaso:
whoops
/lgtm cancel

dprotaso:
@treyhyde quick question - what does your istio installation look like? eg. are you using mesh or just the simple external ingress gateway + cluster local gateway?

dprotaso:
I think the meat of the delegate should just be (meaning we should trim it down to what the activator needs)

```yaml
    route:
    - destination:
        host: hello-00001.default.svc.cluster.local
        port:
          number: 80
      headers:
        request:
          set:
            Knative-Serving-Namespace: default
            Knative-Serving-Revision: hello-00001
```

and then we point the other virtual services to this one.

Note - to handle upgrades I think we should make sure we create the delegate first prior to updating other virtual services

dprotaso:
Oh this block should have the traffic splits too etc

treyhyde:
> @treyhyde quick question - what does your istio installation look like? eg. are you using mesh or just the simple external ingress gateway + cluster local gateway?

Full mesh but the problem would be the same if I was using just a gateway.

treyhyde:
> I think the meat of the delegate should just be (meaning we should trim it down to what the activator needs)
> 
> ```yaml
>     route:
>     - destination:
>         host: hello-00001.default.svc.cluster.local
>         port:
>           number: 80
>       headers:
>         request:
>           set:
>             Knative-Serving-Namespace: default
>             Knative-Serving-Revision: hello-00001
> ```
> 
> and then we point the other virtual services to this one.
> 
> Note - to handle upgrades I think we should make sure we create the delegate first prior to updating other virtual services

I tried to keep the change to just the minimal surface.  I believe it creates the VS with exactly the same function as the other VSs but without the hosts.   If we wanted to do a larger refactoring to base the other VSs on this delegate, we can do that too but it will increase the change surface.

dprotaso:
> I tried to keep the change to just the minimal surface. I believe it creates the VS with exactly the same function as the other VSs but without the hosts. If we wanted to do a larger refactoring to base the other VSs on this delegate, we can do that too but it will increase the change surface.

I think that's fair. Let's just put this change behind a config knob in our config-istio (with the default being off). 

See: https://github.com/knative-extensions/net-istio/blob/main/config/400-config-istio.yaml

Let me know if you have any questions.

dprotaso:
Looks like you'll have to rebase

